PR #37, opened 2021-09-18
Fixed NRE when right clicking a TextBox/AutoCompleteBox

adirh3:
Amazing job with this project so far!
A little fix, after 1.1.2 right clicking `TextBox` or `AutoCompleteBox` cause NRE since `_commandBar` is not created yet.

### DIFF
--- a/FluentAvalonia/UI/Controls/CommandBarFlyout/CommandBarFlyout.cs
+++ b/FluentAvalonia/UI/Controls/CommandBarFlyout/CommandBarFlyout.cs
@@ -163,6 +163,9 @@ namespace FluentAvalonia.UI.Controls
 		protected override void OnOpening(CancelEventArgs args)
 		{
 			base.OnOpening(args);
+			if (_commandBar == null)
+				return;
+			
 			if (PrimaryCommands.Count > 0 && _commandBar.PrimaryCommands.Count == 0)
 			{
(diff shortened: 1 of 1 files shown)


amwx:
Thanks for finding that. Although I'm noticing it now takes two right clicks to get the flyout to appear.

Looking into this occurred, I see why this popped up in 1.1.2. The changes made to `FlyoutBase` for Avalonia 0.10.7 changed the order of events for showing a Flyout. Before `OnOpening` was fired after `CreatePresenter()` which is where the `CommandBar` is created, and now that's reversed, thus it's null. I'm also finding this also isn't the only thing affected by the `FlyoutBase` changes

adirh3:
> Thanks for finding that. Although I'm noticing it now takes two right clicks to get the flyout to appear.
> 
> Looking into this occurred, I see why this popped up in 1.1.2. The changes made to `FlyoutBase` for Avalonia 0.10.7 changed the order of events for showing a Flyout. Before `OnOpening` was fired after `CreatePresenter()` which is where the `CommandBar` is created, and now that's reversed, thus it's null. I'm also finding this also isn't the only thing affected by the `FlyoutBase` changes

Yes, you're right seems like this PR does not fix all issues, so I will be closing it.
I see NREs at `ColorPickerFlyout.OnOpening` due to `FlyoutBase.CancelOpening`.

PR #41, opened 2021-09-24
v1.1.3 Update

amwx:
- Added `DataGrid` styles, note: V2 styles only
- Allow setting MinWidth on `ToggleSwitch`
- Added temporary fix to not have NRE with certain flyouts (CommandBarFlyout & ColorPickerButton). This arose from some changes to FlyoutBase in Avalonia 0.10.7

NOTE:
As of v1.1.3, v1 styles are considered deprecated in FluentAvalonia. You may still continue to use them, however, they will be removed in a future version and new controls within this library, including the NavigationView, will no longer support the original fluent theme. Remaining avalonia controls that do have templates in this library will be added for full compatibility with Avalonia, just with the new styles.

Following the release of Win 11, I will also be making a pass through this library to make sure everything is up to date with WinUI, and adding support for the high contrast theme

### DIFF
--- a/FluentAvalonia/FluentAvalonia.csproj
+++ b/FluentAvalonia/FluentAvalonia.csproj
@@ -10,8 +10,8 @@
 
     <PropertyGroup>
         <PackageLicenseExpression>MIT</PackageLicenseExpression>
-        <Version>1.1.2</Version>
-        <AssemblyVersion>1.1.2.0</AssemblyVersion>
+        <Version>1.1.3</Version>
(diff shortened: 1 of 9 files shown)


robloo:
I think it's a great idea to remove version 1 styles. It's going to be difficult to keep everything in sync anyway and dropping legacy designs will clean a lot up! Version 2 looks a lot better most of the time anyway :)

nlogozzo:
Agreed!

PR #47, opened 2021-10-05
[WIP] v 1.1.5 Update

amwx:
Beginning the 1.1.5 update, which features a full pass over all controls to ensure they're visually ok compared to the new WinUI Fluent v2 styles & matching Windows 11. Just to manage expectations here, this will likely be a long running PR with no ETA at this time. Right now this is mostly serving as my checklist

**New features**:
- Addition of High Contrast theme styles for all controls
- Adds remaining core Avalonia control templates for compatibility with new styles (CaptionButtons, Carousel, ContextMenu, EmbeddableControlRoot, Label, ManagedFileChooser, NotificationCard/WindowNotificationManager, NumericUpDown, TabStrip/TabStripItem, TitleBar)
- Add support to FluentAvaloniaTheme to for new Windows 11 Segoe UI Variable font
- Replace FluentUI Icons font with WinSymbols3 by robloo for compatibility with WinUI Segoe Fluent

**Deprecations/Removals**
- Removal of Fluent v1 theme from FluentAvalonia

[Other updates added as they're completed]

### DIFF
--- a/FluentAvalonia/FluentAvalonia.csproj
+++ b/FluentAvalonia/FluentAvalonia.csproj
@@ -10,8 +10,8 @@
 
     <PropertyGroup>
         <PackageLicenseExpression>MIT</PackageLicenseExpression>
-        <Version>1.1.3</Version>
-        <AssemblyVersion>1.1.3.0</AssemblyVersion>
+        <Version>1.1.5</Version>
+        <AssemblyVersion>1.1.5.0</AssemblyVersion>
     </PropertyGroup>
 
 


robloo:
Will this include use of the newly added CornerRadius to controls? It's useful to update code like this.

https://github.com/amwx/FluentAvalonia/blob/fcfc26f20fc2826c68102e67e0149d052591ce3f/FluentAvalonia/Styling/Controls/ButtonStyle.axaml#L48

Separately, let me know if you need help with, or run across issues, updating to the WinSymbols3 font. 

For example, I can write a quick script to generate https://github.com/amwx/FluentAvalonia/blob/master/FluentAvalonia/UI/Controls/IconElement/Symbol.cs using latest glyph mappings.

I'm not sure if this method is needed at all after this?https://github.com/amwx/FluentAvalonia/blob/fcfc26f20fc2826c68102e67e0149d052591ce3f/FluentAvalonia/UI/Controls/IconElement/SymbolIcon.cs#L108

amwx:
> > Not surprising since I did build the enum off of FluentUI.
> 
> Oh, I thought you were keeping the WinUI Segoe Fluent Unicode points to match upstream. I didn't realize you were building your own symbols different from upstream WinUI.

I meant the when I created the Symbol enum originally. What I actually did was start with the UWP Symbol enum, tried to match what I could of the MDL2 symbols with FluentUI, and then added a bunch of additional icons in there - but it originally was built to fit FluentUI since at the time Segoe Fluent wasn't a thing (or still early) and it and MDL2 have restrictive licenses. 


> Some follow-up questions:
> 
>     1. Is it really a good idea not to match these at all with WinUI upstream? It seems the correct methodology would actually be to use the Segoe Fluent Unicode points, then assign new Unicode points for your new symbols taken from the FluentUI font outside the range used by Segoe Fluent. That means your new symbols would start well above 0xf8b3.

Sorry if I made it sound more confusing than it should be, but that's the intent here - to follow your idea of matching Segoe Fluent where possible. I've already mapped what I could to the Segoe Fluent code points. In the two files I linked above, the values are remapped, and I tested it with WinSymbols3 font and it works. Those enum values that are set to zero are the ones that are missing matching code points, and then in comments I put what I'm currently mapping to.


>     2. With your go-ahead I'll build you a custom font separate from WinSymbols3 that will include new Unicode points. However, I really would keep the existing Segoe Fluent Unicode points in that new font as well for compatiblity.
> 
>     3. With your go ahead I'll fully rebuild the symbols enum to match the new font I propose along with update Unicode points that don't conflict with upstream Windows. I will start in the range U+F0000..U+FFFFD (Supplementary Private Use Area-A).
> 
>     4. The mapping file for your new font will be a .json file. I'll create that in my repository but feel free to copy that here as well.


That would be great! Keep the code points that match the same for compatibility and everything else can be put where ever its not in the way. I don't know how much work this is (and I hope its not too much), but your help on this is greatly appreciated!

robloo:
@amwx Ok, I know how to complete this without too much work -- we are thinking the same now! I'll try to set aside a few hours today/tomorrow to finish it for you.

I will do the following:
 1. Create a new mapping file for new symbols that don't exist in upstream WinUI `FluentAvaloniaSupplemental.json`. These will start in the supplemental private use area-A mentioned above and there should be no overlap with Segoe Fluent Unicode points. This mapping will will ONLY include the symbols that don't exist in Segoe Fluent. **Warning: This will be a breaking change from your current Unicode points: I'll be pushing them much higher in the table so they don't overlap with either FluentUI System or Segoe Fluent.**
 2. Build you a new font using the Segoe Fluent base + merging in your new symbols taken from FluentUI. I'll complete the merge mappings functionality in the SymbolIconManager so I can build a single font from two different mapping files: `SegoeFluent.json` and a new `FluentAvaloniaSupplemental.json`.
 3. I will also merge the enum into a single one for you using the WinUI naming convention 'Fill' instead of 'Filled' suffix. It will include the new Unicode points. I'll try to think of a good way to automate this.

robloo:
I completed the very first build of the mapping file. This generated the JSON file attached below that looks like the given image inside the Symbol Icon Manager. There are errors and I can't build a font from this so there is a fair bit of debugging left I expect. I should have a font by tomorrow though.

This enum is going to be huge -- 505 entries is what it comes out to right now. There are almost 200 in the Windows Symbol enum itself. Then you have added several and included both filled/regular variants. This also doesn't include all the other icons that will be merged in from the standard SegoeFluent font. When complete:

 1. FluentAvalonia.ttf will include all Unicode points available from the WinSymbols3 (Segoe Fluent Icons)
 2. FluentAvalonia.ttf will include all Unicode points from the Symbol enum that correspond with the very old Segoe UI Symbol font from Windows 8 (this enum is the only place they are still used).
 3. FluentAvalonia.ttf will include all of the new regular/filled icons you've added from the Fluent UI System
 4. I will also generate a new Symbols enum for you from this mapping file.

After the initial mapping file is complete, edits should be made within the JSON file directly or through the UI in the Symbol Icon Manager. There is a very complicated build process to generate the initial mapping file but it isn't needed going forward.

image

[FluentAvalonia.zip](https://github.com/amwx/FluentAvalonia/files/7351026/FluentAvalonia.zip)

amwx:
I've only taken a quick look, but overall it looks like it's working really well. One observation: Were there some new symbols added to the enum, some of them I don't remember being present? That's fine if so, I just want to make sure I'm not going crazy.
There also seems to be some symbols missing, but I'll take a closer look at that later.

In related news, as it stands, the new font file will cut ~2MB off the package/solution size, so that's awesome

robloo:
>Were there some new symbols added to the enum, some of them I don't remember being present?

Yes, I wanted the enum to be 100% compatible with upstream so it includes ALL values from the WinUI [Symbol enum](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.symbol?view=winrt-22000). Then I also added in all of your new ones as well. 505 in total.

The font itself includes everything in the enum plus everything in SegoeFluent that exists in the current mapping file. This means the FluentAvalonia font is also compatible with SegoeFluent while including your new symbols as well (in much higher Unicode points so there is no overlap).

The initial build process was more complicated than I would have liked but you can read through it starting here: https://github.com/robloo/SymbolIconManager/blob/acceda83c61b2eb3d26624319ac47678c7e4ebba/Source/Models/Specialized/FluentAvalonia.cs#L33

> There also seems to be some symbols missing, but I'll take a closer look at that later.

The enum includes all symbols in WinUI and not all of them have an equivalent in the new font. I just don't know a good mapping for them. This can be improved by updating the SegoeFluent.json mapping file and then rebuilding all fonts. I decided it's better for the enum to have an entry even if the glyph doesn't exist (so copy/pasted code compiles).

In addition, I noticed the conversion to size 20 FluentUISystem SVGs doesn't seem to be working everywhere so I definitely will give the font another look. Just let me know what else you might find and I'll rebuild it next week.

> In related news, as it stands, the new font file will cut ~2MB off the package/solution size, so that's awesome

Perfect! Another bonus of the Symbol Icon Manager is can build custom fonts specifically for an app. This is for the same reason: to cut down on app package size. This is especially helpful for some larger apps.

amwx:
> The font itself includes everything in the enum plus everything in SegoeFluent that exists in the current mapping file. This means the FluentAvalonia font is also compatible with SegoeFluent while including your new symbols as well (in much higher Unicode points so there is no overlap).
> The enum includes all symbols in WinUI and not all of them have an equivalent in the new font. I just don't know a good mapping for them. This can be improved by updating the SegoeFluent.json mapping file and then rebuilding all fonts. I decided it's better for the enum to have an entry even if the glyph doesn't exist (so copy/pasted code compiles).

Perfect. What I'll probably do is just mark the missing WinUI symbols with an `[Obsolete]` attribute so they keep their position, but it's known don't use them. Then they can be added back in the future if something becomes available.

I've given everything a closer look, and there's only a few missing that should be there:

These show a mapping, but are showing the 'invalid glyph' character.
CloudBackup/CloudBackupFilled
CloudDownload/CloudDownloadFilled
CloudOffline/CloudOfflineFilled
CloudSyncComplete/CloudSyncCompleteFilled
DockLeftFilled
DockRightFilled
Restore/RestoreFilled
WifiProtected/WifiProtectedFilled


These should be showing something, but don't have a mapping present. These have a code point for the 'Destination' but is undefined for the source.
DockLeft (0xFBF8 in FluentUI)
DockRight (0xFBFD in FluentUI)
DockBottom (0xFF78 in FluentUI)
FontColor (0xF7BF in FluentUI)
MailReply (0xF177 in FluentUI)
PreviewLink (0xF625 in FluentUI)
Scan (0xF68B in FluentUI)

How difficult is it to build the font itself now that the mapping file is generated? Good to know for future use, and if it's not hard I could probably fix the second set above myself.

robloo:
Yea, what you found with the first set matches with the log file. The build script wasn't able to locate a few SVG files. I'm not sure why but I'll take a look.

The second set is probably because those mappings are missing in SegoeFluent.json (the mapping to build that font). Those mappings are master and will overwrite anything in the custom FluentAvalonia mappings file. I shouldn't be overwriting a good source with a missing one though so that's a bug in merge functionality I'll fix. I'll also update SegoeFluent font itself. 

I'll rebuild the font again from initial after fixing this. Don't worry about updating the json mapping file. Afterwards, I'll document how to font, its quite easy on Windows. I'll also probably make a button for known fonts to make it even easier.

robloo:
@amwx 

Can you please re-review the below list in upstream FluentUISystem repo: https://github.com/microsoft/fluentui-system-icons. It appears they have removed them entirely from the font (they do not exist in the upstream .ttf files and I can't locate the SVG files either). SVG files are critical in order to build a font. They also don't exist anymore in https://github.com/microsoft/fluentui-system-icons/blob/master/icons.md. The only place they are still referenced in [FluentSystemIcons-Filled.json](https://github.com/microsoft/fluentui-system-icons/blob/master/fonts/FluentSystemIcons-Filled.json) and [FluentSystemIcons-Regular.json](https://github.com/microsoft/fluentui-system-icons/blob/master/fonts/FluentSystemIcons-Regular.json) which is why they still are detected and processed by the Symbol Icon Manager.

Perhaps they changed the name and didn't properly update the repository: cloud_backup could now be cloud_archive for example (I can't compare images to tell though). I'm not sure what to do here.

 1. It might be easiest to remove them from FluentAvalonia and your enum
 2. We can try to find equivalent replacements (assuming they were just incorrectly renamed)
 2. We can go back in the FluentUISystem repository history and save a copy of the SVGs before they were removed. Then rebuild the font instead using custom images. (For example the source would be IconSet=Undefined, Name="ic_fluent_cloud_backup_48_regular" UnicodePoint=0)

### Your list
CloudBackup/CloudBackupFilled
CloudDownload/CloudDownloadFilled
CloudOffline/CloudOfflineFilled
CloudSyncComplete/CloudSyncCompleteFilled
DockLeftFilled
DockRightFilled
Restore/RestoreFilled
WifiProtected/WifiProtectedFilled

### Build Log
Error: Missing SVG source URL, mapping skipped src=0xF2E8, dst=0xF8039 (ic_fluent_cloud_backup_48_regular)
Error: Missing SVG source URL, mapping skipped src=0xF2E8, dst=0xF803A (ic_fluent_cloud_backup_48_filled)
Error: Missing SVG source URL, mapping skipped src=0xFE5B, dst=0xF803C (ic_fluent_cloud_download_20_filled)
Error: Missing SVG source URL, mapping skipped src=0xF2EC, dst=0xF8040 (ic_fluent_cloud_offline_24_regular)
Error: Missing SVG source URL, mapping skipped src=0xF2EC, dst=0xF8041 (ic_fluent_cloud_offline_24_filled)
Error: Missing SVG source URL, mapping skipped src=0xF2EE, dst=0xF8043 (ic_fluent_cloud_sync_complete_48_regular)
Error: Missing SVG source URL, mapping skipped src=0xF2EE, dst=0xF8044 (ic_fluent_cloud_sync_complete_48_filled)
Error: Missing SVG source URL, mapping skipped src=0xFC03, dst=0xF8060 (ic_fluent_dock_panel_left_48_filled)
Error: Missing SVG source URL, mapping skipped src=0xFC08, dst=0xF8061 (ic_fluent_dock_panel_right_48_filled)
Error: Missing SVG source URL, mapping skipped src=0xF670, dst=0xF80B9 (ic_fluent_restore_16_regular)
Error: Missing SVG source URL, mapping skipped src=0xF67A, dst=0xF80BA (ic_fluent_restore_16_filled)
Error: Missing SVG source URL, mapping skipped src=0xF8B4, dst=0xF812A (ic_fluent_wifi_protected_24_regular)
Error: Missing SVG source URL, mapping skipped src=0xF8CC, dst=0xF812B (ic_fluent_wifi_protected_24_filled)

robloo:
Ah, I found the commit that seems to have changed all these. The good news is that they likely were just renamed.

https://github.com/microsoft/fluentui-system-icons/pull/331

amwx:
In terms of fixing the above problems, I think we're good. The only thing that was mis-matched was 'Target' which is mapping to RadioButtonFilled from FluentUI instead of '_target_20'

However, I think I just found a slightly larger problem (unless you've already done this). I don't think WinUI ever uses a SymbolIcon in any of its controls, preferring to map directly to the glyph. So other glyphs not in the Symbol enum are now missing. Biggest example is the little carets used in the ScrollBars and the CalendarView. I'm now wondering if we need to match Segoe Fluent font to accommodate this. They wouldn't need to be in the enum (that's got enough as it is), but will at least need to be in the font file for full compatibility.

robloo:
> The only thing that was mis-matched was 'Target' which is mapping to RadioButtonFilled from FluentUI instead of '_target_20'

Thanks, I'll be sure to fix that. It's likely an issue with the SegoeFluent mappings.

> I don't think WinUI ever uses a SymbolIcon in any of its controls, preferring to map directly to the glyph.

Yes, correct. However, WinUI also defines the resource to use for this `{ThemeResource SymbolThemeFontFamily}`. This means you should only need to define a new font here and it should be updated everywhere. This is what the Uno Platform does and it works quite well. We even made sure upstream is fixed to never hard-code Segoe MDL2 Assets https://github.com/microsoft/microsoft-ui-xaml/issues/3745.

That said, Microsoft is using some old codepoints that they shouldn't: https://github.com/microsoft/microsoft-ui-xaml/issues/5887. So a few glyphs won't appear even with the new font. I should probably support these old codepoints in the SegoeFluent replacement font (WinSymbols3) but don't because they really should have been obsoleted by now. 

I would suggest for Fluent Avalonia you update to the latest codepoints. I have all the mappings from the old, original Segoe UI Symbol codepoints to the new ones in SegoeMDL2Assets/SegoeFluent here: https://github.com/robloo/SymbolIconManager/blob/main/Source/Data/Mappings/SegoeUISymbolToSegoeMDL2Assets.json. It would just be a find-replace and things should work.

If all of this still doesn't get things to work, I never defined a name in the font file so maybe that won't allow you to select the font properly in XAML. You'll need to do something like: `"/Assets/Fonts/SourceSansPro-Italic.ttf#Source Sans Pro"` I expect.

> I'm now wondering if we need to match Segoe Fluent font to accommodate this. They wouldn't need to be in the enum (that's got enough as it is), but will at least need to be in the font file for full compatibility.

I'm already doing this mentioned above. "The font itself includes everything in the enum plus everything in SegoeFluent that exists in the current mapping file. This means the FluentAvalonia font is also compatible with SegoeFluent while including your new symbols as well (in much higher Unicode points so there is no overlap)."

amwx:
@robloo  Thanks. I got myself confused between still having the old code points but changing the font. I already have `SymbolThemeFontFamily` declared, now I just have to go back through and change all the code points from FluentUI to the new font which matches WinUI, and switching the fonts should be complete! Huge thanks for all your help.

robloo:
Let's hope this does it!

[FluentAvalonia_v3.zip](https://github.com/amwx/FluentAvalonia/files/7369824/FluentAvalonia_v3.zip)

There are lots of fixes and synchronizations with upstream SegoeFluent. Major changes include:

- Fixed Target glyph
- Fixed Wifi glyphs that were incorrect, synced with SegoeFluent as well
- Corrected build to automatically use size 20 where possible (was a stupid mistake and an easy fix!)
- Where a name matches with SegoeFluent, SegoeFluent's Unicode point is now used instead of a new, custom one in some cases.

Note that these changes mean everything is updated including the enum as there were Unicode point changes.

I'm pretty happy where things are with the build process even if it's a lot more complicated than I would like. It is also fully automated now. Feel free to browse through the code if you get a chance in the future and you should be able to make updates yourself as needed without too much effort.

 * [Mapping file build/rebuild process](https://github.com/robloo/SymbolIconManager/blob/05f4fbe2db9e1ded0a40aedb3219f885de8326da/Source/Models/Specialized/FluentAvalonia.cs#L41)
 * [FluentAvalonia.ttf build](https://github.com/robloo/SymbolIconManager/blob/05f4fbe2db9e1ded0a40aedb3219f885de8326da/Source/Views/FontsView.axaml.cs#L257-L281). There is now only 1 button to click and out comes your font!

Issue #22, opened 2021-07-21
[Feature Request] Update TabView


nlogozzo:
Microsoft with new WinUI has created a new `TabView` that replaces `TabControl` (https://docs.microsoft.com/en-us/windows/apps/design/controls/tab-view) The old `TabControl` is known as a `Pivot` (https://docs.microsoft.com/en-us/windows/apps/design/controls/pivot)
Could we see these updated controls in this library?

amwx:
I wouldn't necessarily say it's a replacement for `TabControl`/`Pivot` since it's a heavier weight, top-level navigation focused control - and why so many complained when MS tried to drop the `Pivot` control. I will look into adding it though as there is value to it, but no timeline for it

nlogozzo:
No worry about the timeline, I like the current TabControl/Pivot navigation...however the TabView would be helpful for some future projects

nlogozzo:
Any update on this? @amwx

amwx:
Its on my list of things to add after v1.2, which I'm working on right now - that is taking longer than I'd like because it's required reworking some things. v1.2 will be out sometime in December, so probably early next year time frame wise for the TabView. The good news is TabView is open sourced so it should be relatively easy, and I've peeked at the source code and some of the stuff I'm currently working on may be needed for the TabView, so it sets itself up nicely.

nlogozzo:
Sounds good! Thanks!!

PR #72, opened 2022-01-18
Codebase cleanup, changes: v1.2 part 1

amwx:
Beginning of the v1.2 update, which will feature primarily codebase cleanup, optimizations, and the like - which yes, is a little bit of a different scope from what I was originally planning. I've realized, while continuing work on the ListView/GridView ports that I started, that I've stretched myself too thin and been over ambitious with my goals. So, to alleviate that, I've suspended that work for the time being - I want to re-evaluate how I'm going about that, including experimenting if rewriting the ListView to work on the ItemsRepeater is a better alternative - since getting virtualization right is proving to be a huge challenge. Though the ItemsRepeater does have it's issues, and I'm not sure how it would work, it does work really nicely most of the time. I also noticed there was some work started on a potential new ItemsControl upstream in Avalonia, which may also work. 

v1.2 is broken up into multiple parts to help me out, but to better outline the goals. This PR is the largest and focuses on some codebase cleanup and optimizations I've been wanting to do for a while and really need to be done now before more controls are added.

[General changes]
- All public facing classes, properties, and methods will now have XML documentation. Along with enums, where necessary.
- I've removed any unused items within the Win32Interop.cs file. FluentAvalonia started as my personal control library which is why that was there - but it's unnecessary to keep most of the stuff there.
- EventArgs classes have been moved to their own file where needed, and where appropriate, enums too.
- Added an .editorconfig file
- Added a few unit tests - a slow, but ongoing process

[UI.Controls]
- Most controls will now have their AvaloniaProperties, CLR Properties, events, and property backing fields in a separate file (identified as `[ControlName].properties.cs`) to improve clarity. Controls whose class size is relatively small will be exempt, however. For example, `HyperlinkButton` defines 1 property and doesn't have much logic - so will remain as a single class.
- Clean up of formatting and comments in various controls
- `ColorPickerButton`, `SplitButton`, and `ToggleSplitButton` have been moved to their own folder
- Removed the `ItemsSourceView` within the NavigationView folder and change to use the upstream one in Avalonia. (shouldn't result in any noticable differences or breaking changes)
- As stated in 1.1.7, cases where WinUI events use `object` have will now use `EventArgs`. 1.1.7 changed from passing `null` to `EventArgs.Empty` but didn't change the event handler because it is a breaking change. This change is now made

[ColorPicker]
- Changed to use `ImmutablePen` in the component controls
- Dispose of the event handlers in `ColorPicker` to prevent leaks in ApplyTemplate

[NumberBox]
- Changed some properties from DirectProperty to StyledProperty
- Used AddOwner<>() where appropriate

Part 2 will likely be restructuring some of the Styles and FluentAvaloniaTheme. The extent of this is TBD, as I need to figure out what I can actually do without needing xaml compiler support. 
Part 3 will be a new sample app


nlogozzo:
Keep up the great work, really excited for this update!! (hope to see #22 as well :smile: )

amwx:
> Keep up the great work, really excited for this update!! (hope to see #22 as well 😄 )

I am going to do my best to get that in, if not, it will be the first thing I do next. Since it uses WinUI's ListView internally, I'll need see how I can build it without that and use the existing ListBox instead - which is doable and not hard, just tedious.

nlogozzo:
> > Keep up the great work, really excited for this update!! (hope to see #22 as well smile )
> 
> I am going to do my best to get that in, if not, it will be the first thing I do next. Since it uses WinUI's ListView internally, I'll need see how I can build it without that and use the existing ListBox instead - which is doable and not hard, just tedious.

No rush, thanks!

PR #83, opened 2022-01-25
FluentAvaloniaTheme updates v1.2, part 2

amwx:
This PR is a major change to the way FluentAvaloniaTheme works and features several breaking changes:

- Use MicroCom WinRT interop and reduce amount of undocumented API calls used for Windows system resources
- Changes to resources:
--`LightResources.axaml`, `DarkResources.axaml`, and `BaseResources.axaml` are now `ResourceDictionary` types, rather than `Styles`
--AccentColors, ContentControlThemeFontFamily, and SymbolThemeFontFamily are now defined inside FluentAvaloniaTheme and not exposed in any resource dictionaries (they still work exactly the same, however).
--FluentAvaloniaTheme now searches App level resource dictionary (App.axaml) before it searches the Fluent theme resources allowing you to override values now. Note that the app level styles collection is not searched, only the resource dictionary (App.Resources)
- *New Feature*: High Contrast theme will now correctly resolve the colors from Windows (and the variants on Windows 11)
- **breaking change** Properties have been renamed:
-- `UseSegoeUIOnWindows` is now `UseSystemFontOnWindows`
-- `GetUserAccentColor` is now `UseUserAccentColorOnWindows`
-- `DefaultToUserTheme` is now `UseSystemThemeOnWindows`
-- **breaking change** Method renamed
-- `ForceNativeTitleBarToTheme` is now `ForceWin32WindowToTheme` - this change better reflects that it's win32 only and applies not only to the titlebar, but also the window border. 
- NEW: `SkipControls` property added: this property is a semicolon delimited string that will let you skip certain controls when templates are loading. For example, if you never use the DataGrid or the NavigationView, you can specify those controls here and the templates will not be loaded when FluentAvaloniaTheme initializes. This is a startup value only and more on how this works will come later...
- In order to allow the above to work, `Controls.axaml` was removed and a list of all control templates is provided in text file format, which is parsed at load
- ~**breaking change** `CustomAccentColor` property removed as you can now override SystemAccentColor and its variants directly in App.axaml. However, you will need to generate the light/dark shades yourself.~ 
EDIT: `CustomAccentColor` property will be kept
- **breaking change** `RequestedThemeChanged` event now has it's own event args class instead of just passing a string. The new theme can be retrieved from the event args.
- All control styles files have been updated to end with "Styles", e.g., (ButtonStyles.axaml) for consistency. **Breaking change** only if you reference styles files individually and don't use FluentAvaloniaTheme, otherwise, this is already handled


nlogozzo:
I feel as if you should keep `CustomAccentColor` as in my applications I have the ability for the user to select from a list of accent colors for customization.

amwx:
> I feel as if you should keep `CustomAccentColor` as in my applications I have the ability for the user to select from a list of accent colors for customization.

Note that I'm not removing the ability to set a custom accent, just transferring the responsibility. Now that you can override resources at the app level, you can provide the accents there. The downside is that you'll have to generate the light/dark variants yourself, but that's trivial. However, this now gives anyone full access to how they want the variants generated. 

If you want to supply custom color, add to App level resource dictionary
`Application.Current.Resources.Add("SystemAccentColor", myColor);`
or update via
`Application.Current.Resources["SystemAccentColor"] = myUpdatedColor;`
& repeat for variants

If you want to return to default, just remove them (note that the remove feature on AvaloniaDictionary isn't functional and a recent PR fixed that)

nlogozzo:
> > I feel as if you should keep `CustomAccentColor` as in my applications I have the ability for the user to select from a list of accent colors for customization.
> 
> 
> 
> Note that I'm not removing the ability to set a custom accent, just transferring the responsibility. Now that you can override resources at the app level, you can provide the accents there. The downside is that you'll have to generate the light/dark variants yourself, but that's trivial. However, this now gives anyone full access to how they want the variants generated. 
> 
> 
> 
> If you want to supply custom color, add to App level resource dictionary
> 
> `Application.Current.Resources.Add("SystemAccentColor", myColor);`
> 
> or update via
> 
> `Application.Current.Resources["SystemAccentColor"] = myUpdatedColor;`
> 
> & repeat for variants
> 
> 
> 
> If you want to return to default, just remove them (note that the remove feature on AvaloniaDictionary isn't functional and a recent PR fixed that)
> 
> 
> 
> EDIT: After some thinking, I've decided to go ahead and keep this property. Both options are now available, you can set `CustomAccentColor` and get the same behavior as before, or you can directly override them to your liking.
> 
> 
> 
> 

Thanks for keeping the property :) looking forward to this update!

PR #84, opened 2022-01-29
NumberBox - Added ValidationErrors to ValueProperty

RobertBeekman:
This PR enables validation errors in the NumberBox control.

image

I took extra care not to break the placement of the spinner in both compact and inline.

I'm sorry about the changes in tabs/spaces in axaml files. I couldn't for the life of me wrestle my IDE into matching the existing pattern.


amwx:
Great! I still don't understand why the data errors aren't automatically inherited to the `TextBox` to handle it, but whatever. No worries on the IDE gone crazy with the spacing.

Also FYI, all the current sample pages are being scrapped as part of my overhaul of the Sample App, but I will try to copy over your validation example into the new page.

amwx (approved):

RobertBeekman:
Good stuff, don't worry about the sample app I just threw it in there to make sure it actually works with validation errors :)

PR #98, opened 2022-02-17
[ContentDialog] Fix height issue if bigger than window

### LINKED ISSUE #96: ContentDialog buttons are hidden with large content
**Describe the bug**
I'm using a custom content in ContentDialog, it is a ListBox with several items, but the buttons is hidden in window,  even when I scroll to the end of the scrollbar. The screenshot is my window with the ContentDialog shown, the scrollbar is at the end

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Windows 10
 - FluentAvalonia Version 1.2.1
 - Avalonia Version 0.10.12.1

amwx:
Fixes #96 

Small fix to the ContentDialog to fix an issue if the Dialog measures larger than the window to ensure the dialog is never bigger than the window. Fixed by measuring the child of DialogHost with the constrained size of the VisualRoot client size rather than infinity

### DIFF
--- a/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
+++ b/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
@@ -1,10 +1,12 @@
 ﻿using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Controls.Metadata;
 using Avalonia.Controls.Presenters;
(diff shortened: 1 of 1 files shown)


Maruhl:
I would try to make the dialog not larger than the area of the screen where the main window is. Because if the main window is too small, the dialog is also so small, but some dialogs have so much content that this again cannot be displayed completely. Ergo, the user must first maximize the main window to be able to operate the dialog at all - which is again too many mouse clicks -> cumbersome operation.

PR #101, opened 2022-02-20
Added flyout events to color picker button

RobertBeekman:
Hi there,

The ColorPickerButton had no way of letting me know when it's flyout was being opened or closed, so I added those events in this PR.

For my own application I've been using the color picker with the confirm/dismiss buttons disabled to realtime show the difference your new selection makes (the new color is displayed on the LEDs of devices). This is great but I need to know when the user is finished picking a color so I can execute an undo/redo command, hence this PR :)

I made a slight tweak to `ColorPickerFlyout` here to make sure events happen in a (to me) more logical order, it also made managing the `_isFlyoutOpen` state of `ColorPickerButton` a bit easier.

The order of things is now always:
- `OnFlyoutOpened`
- `OnFlyoutDismissed` or `OnFlyoutConfirmed`
- `OnFlyoutClosed`

Ofcourse the dismissed/confirmed events never fire if you disabled that functionality, I made sure this doesn't break anything.

PS: I really like what you did with the demo application, great job 👍🏻

PR #108, opened 2022-03-01
[WIP] FluentAvalonia v1.3

### LINKED ISSUE #100: Broken DataGrid Styles
The spacing between DataGrid columns seems to be broken:
image

DataGrid defined as follows:
```xml
<DataGrid Margin="6,6,6,6" MinHeight="200" AutoGenerateColumns="False" IsReadOnly="True" SelectionMode="Extended" GridLinesVisibility="Horizontal" Items="{Binding Downloads}">
  <DataGrid.Columns>
    <DataGridTextColumn Header="Filename" Binding="{Binding FullFilename}" MinWidth="240"/>
    
    <DataGridTextColumn Header="Video Link" Binding="{Binding VideoLink}" MinWidth="300"/>
    
    <DataGridTextColumn Header="Status" Binding="{Binding Status}" MinWidth="200"/>
    
    <DataGridTemplateColumn Header="Progress" MinWidth="200">
	    <DataGridTemplateColumn.CellTemplate>
		    <DataTemplate>
			    <Grid>
				    <ProgressBar HorizontalAlignment="Stretch" VerticalAlignment="Stretch" Height="10" Minimum="0" Maximum="10" Value="{Binding Progress}"/>
			    </Grid>
		    </DataTemplate>
	    </DataGridTemplateColumn.CellTemplate>
    </DataGridTemplateColumn>
  </DataGrid.Columns>
</DataGrid>
```

amwx:
Quite a large update we have here, two new controls and a bunch of fixes:

**New Controls**
- Added `TabView` control
NOTE: As of now, `TabView` does not support HighContrast styles - they are disabled. This is because the WinUI team still uses the old Fluent v1 naming in those styles - which breaks here when those names don't exist. 
- Added `TaskDialog`s

These controls and the below fixes are available now in the v1.3.0-beta2 nuget package

**Fixes/Improvements**
- Fixed issue where icons in NumberBox popup inside ColorPicker would show invalid glyphs (#95)
- Fixed issue where ContentDialog wouldn't size correctly if window height was smaller than dialog height (#98)
- Flyout events added to ColorPickerButton (#101, thanks to RobertBeekman)

*`Frame`*
- Added support for the `NavigateToType(Type, object, FrameNavigationOptions)` method
- Added support for saving and restoring the Frame's history (`GetNavigationState`/`SetNavigationState`) - currently only save Page type and parameter (if it can be serialized into a string). 
- Added RoutedEvents to allow passing a parameter to the pages that are being Navigated to
Since we don't have a `Page` control, the methods `OnNavigatedTo()`, `OnNavigatingFrom()`, and `OnNavigatedFrom()` don't exist, and passing a parameter to the pages is quite difficult. You can now handle the events to retreive the navigation information when navigating between pages in your app. Notice the routing strategy is `Direct`, it is only posted to the page being loaded/unloaded - and does not bubble/tunnel through the tree
```C#
AddHandler(Frame.NavigatingFromEvent, Handler, RoutingStrategies.Direct);
AddHandler(Frame.NavigatedFromEvent, Handler, RoutingStrategies.Direct);
AddHandler(Frame.NavigatedToEvent, Handler, RoutingStrategies.Direct);
```
These changes just about complete `Frame` and bring it up to what it is in WinUI.

*`IconSource`*
- Fixed issue where setting the Foreground on an IconSource wouldn't be respected in the actual IconElement

*`DataGrid`*
- Fix broken DataGrid TextBlock column margin from upstream (fixes #100)

*`FluentAvalonia.UI.Controls.MenuFlyout` and related*
- Fixed weirdness with `RadioMenuFlyoutItem`
- Enabled two-way binding by default on `RadioMenuFlyoutItem` and `ToggleMenuFlyoutItem`
- Added `InputGesture` property to `MenuFlyoutItem` to enable adding a key gesture without needing a HotKey. This is essentially the equivalent of WinUI's `KeyboardAcceleratorTextOverride` property
- Fixed issue where moving the mouse over a `MenuFlyoutSubItem` and then to another menu item would cause the sub menu to temporarily open
- Fixed Keyboard Navigation* in a MenuFlyout
*=there's more to be done, but this is more related to what I'm looking into for re-writing Avalonia's FocusManager and navigation support

Still TODO:
- [ ] Fix some issues on the `TabView` sample page
- [ ] Allow expanding/collapsing the Xaml/Csharp source examples in the SampleApp

After v1.3:
Last I heard Avalonia is planning v11.0 for sometime in April - whether the current world chaos will change that timeline is unknown. However, given that v11.0 will have some significant changes (new renderer, styling system changes, etc.), I'm looking to see about getting a branch that reflects Avalonia's master branch and begin making incremental changes along the way, as I'm expecting things to break. I've applied for a free myget package feed, but they haven't gotten back to me yet, so IDK...

New controls will probably be held off until Avalonia v11.0, unless something is super pressing, to limit the overall work needed to upgrade to the major version changes

### DIFF
--- a/FASandbox/FASandbox.csproj
+++ b/FASandbox/FASandbox.csproj
@@ -7,10 +7,10 @@
     <None Remove=".gitignore" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Avalonia" Version="0.10.12" />
-    <PackageReference Include="Avalonia.Desktop" Version="0.10.12" />
+    <PackageReference Include="Avalonia" Version="0.10.13" />
+    <PackageReference Include="Avalonia.Desktop" Version="0.10.13" />
     <!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
-    <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="Avalonia.Diagnostics" Version="0.10.12" />
+    <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="Avalonia.Diagnostics" Version="0.10.13" />
    </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\FluentAvalonia\FluentAvalonia.csproj" />
--- a/FluentAvalonia/FluentAvalonia.csproj
+++ b/FluentAvalonia/FluentAvalonia.csproj
@@ -5,6 +5,8 @@
     <Nullable>disable</Nullable>
 	<RepositoryUrl>https://github.com/amwx/FluentAvalonia</RepositoryUrl>
 	<PackageId>FluentAvaloniaUI</PackageId>
+    <Description>Control library focused on fluent design and bringing more WinUI controls into Avalonia </Description>
+    <PackageTags>c-sharp;xaml;cross-platform;dotnet;dotnetcore;avalonia;avaloniaui;fluent;fluent-design</PackageTags>
     <MicroComGeneratorRuntimeNamespace>MicroCom.Runtime</MicroComGeneratorRuntimeNamespace>
 </PropertyGroup>
 
@@ -18,8 +20,8 @@
 
     <PropertyGroup>
         <PackageLicenseExpression>MIT</PackageLicenseExpression>
-        <Version>1.3.0-beta1</Version>
-        <AssemblyVersion>1.2.1.999</AssemblyVersion>
+        <Version>1.3.0</Version>
+        <AssemblyVersion>1.3.0.0</AssemblyVersion>        
     </PropertyGroup>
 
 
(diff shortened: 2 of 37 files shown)


nlogozzo:
Would it be hard for you to implement a `Page` control in the library?

amwx:
> Would it be hard for you to implement a `Page` control in the library?

Probably not. But...it's another thing to maintain. And there's no template for it in VS/Rider, so you'd have to manually change a UserControl over to Page. Not everyone will do that, which means Frame would need to adapt. I think what I came up with is a good solution, really the only difference is instead of having the 3 override methods, you just add an event handler for which event you want.

PR #114, opened 2022-03-13
Update FluentAvalonia.ttf

robloo:
Fixes https://github.com/robloo/SymbolIconManager/issues/1 first discussed in https://github.com/amwx/FluentAvalonia/pull/104#issuecomment-1050534958

The font was rebuilt and the metrics were manually adjusted.


amwx (approved):
Thanks again for doing this! As an added bonus, some of the aliasing issues have also been fixed with the alignment corrections too (the middle line on hamburger navigation icon, for ex).

PR #124, opened 2022-04-04
Use AddOrUpdateSystemResource as the reseocue may already be set

timunie:
This is a fix for an issue where one cannot change the Theme to the CurrentSystemTheme after startup 

If you need more info or a ticket first, let me know

Happy coding
Tim

### DIFF
--- a/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
+++ b/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
@@ -391,8 +391,11 @@ namespace FluentAvalonia.Styling
             {
                 _requestedTheme = newTheme;
 
-                // Remove the old theme resources
-                _themeResources.MergedDictionaries.RemoveAt(1);
+                // Remove the old theme if any resources
+                if (_themeResources.Count > 0)
+                {
+                    _themeResources.MergedDictionaries.RemoveAt(1);
+                }
 
                 _themeResources.MergedDictionaries.Add(
                     (ResourceDictionary)AvaloniaXamlLoader.Load(new Uri($"avares://FluentAvalonia/Styling/StylesV2/{_requestedTheme}Resources.axaml"), _baseUri));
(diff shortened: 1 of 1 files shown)


amwx:
I'm away right now, so it won't be until probably this weekend that I can look at this more, but what was the exact issue that was leading to the exception being thrown?

timunie:
Don't worry. I opened a ticket with full stack trace. btw: I think this is also totally valid for adding or updating:

```cs
Resources["MyKey"] = "Abc"; // Add
Resources["MyKey"] = "Def"; // Update
```

Have a great time
Tim

timunie:
fyi: Found another potentially bug which I pushed a fix for also.

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
```diff
@@ -383,16 +383,19 @@ private void Refresh(string newTheme)
...
 
             var old = _requestedTheme;
             if (!string.Equals(newTheme, old, StringComparison.OrdinalIgnoreCase))
             {
                 _requestedTheme = newTheme;
 
-                // Remove the old theme resources
-                _themeResources.MergedDictionaries.RemoveAt(1);
+                // Remove the old theme if any resources
+                if (_themeResources.Count > 0)
```
amwx, line 395: There should always be 2 dictionaries within `_themeResources.MergedDictionaries`. In what case were you finding that not to be?

timunie: It happens when you request a special Theme in App.axaml. I guess then the Theme wasn't loaded, so nothing to remove

In the demo app try 

```xml
<style:FluentAvaloniaTheme RequestedTheme="Light" />
```

amwx: Ahh ok, yep. The root cause here is that setting `RequestedTheme` calls `Refresh()`, which can happen before everything initializes. We can keep the check you put in here since it doesn't hurt anything, but can you adjust the setter of `RequestedTheme` to:

```C#
set
{
    _requestedTheme = value;
    if (_hasLoaded)
       Refresh(value);
}
````

This will defer loading a new theme if we haven't initialized yet and will prevent that exception.

timunie: good point 👍 

I've updated the PR now.

amwx (approved):

PR #136, opened 2022-04-28
Fix for PathIcon wrong aspect ratio 

### LINKED ISSUE #135: PathIcon gets stretched if the input geometry is not square
**Describe the bug**
I have icons that do not fit in a uniform box. this leads to different scalings in X and Y direction. That makes these icons look ugly IMO. 

Example: 
```xml
<UserControl.Resources>
    <StreamGeometry x:Key="MyIcon">M 0 0, 0 4, 8 2 z</StreamGeometry>
</UserControl.Resources>

<ui:PathIcon Data="{DynamicResource MyIcon}" 
             Width="50" Height="50" />
```

**Screenshots**

This is how the icon should look like: 

image

This is the actual result:

image

Don't care about the different scaling, but mind the aspect ratio. 

Another Example: 

image
image

 - OS: [Windows (10)]
 - FluentAvalonia Version [latest]
 - Avalonia Version [0.10.13]

**Additional context**
I think I have an idea where this issue comes from. If I manage to fix it, I'll provide a PR for this.

timunie:
Hi @awmx, 

this PR should close #135 . I hope that this does not break anything, but I can't promise. Someone may want to get the icon streched and not take aspect ratio into account. 

Fixes #135 

Result: 
image

Happy coding
Tim

### DIFF
--- a/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
+++ b/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
@@ -1,4 +1,5 @@
-﻿using Avalonia;
+﻿using System;
+using Avalonia;
 using Avalonia.Media;
 using System.Collections.Generic;
 
@@ -50,10 +51,19 @@ namespace FluentAvalonia.UI.Controls
             //the bounds specified
             var bounds = data.Bounds;
             var destRect = new Rect(Bounds.Size);
-            var scale = Matrix.CreateScale(
-                destRect.Width / bounds.Width,
-                destRect.Height / bounds.Height);
-            var translate = Matrix.CreateTranslation(-(Vector)bounds.Position);
+
+            // Prevent divide by zero exception
+            if (bounds.Width == 0 || bounds.Height == 0 || destRect.Width == 0 || destRect.Height == 0) return;
+            
+            var scaleFactor = Math.Min(destRect.Width / bounds.Width, destRect.Height / bounds.Height);
+            
+            var offsetCenter = new Vector(
+                (bounds.Width - destRect.Width / scaleFactor) / 2,
+                (bounds.Height - destRect.Height / scaleFactor) / 2);
+            
+            var scale = Matrix.CreateScale(scaleFactor, scaleFactor);
+
+            var translate = Matrix.CreateTranslation(-(Vector)bounds.Position - offsetCenter);
 
             using (context.PushClip(destRect))
             using (context.PushPreTransform(translate * scale))
--- a/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
+++ b/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
@@ -1,7 +1,9 @@
-﻿using System;
+using System;
 using Avalonia;
 using Avalonia.Media;
 using System.Collections.Generic;
+using Avalonia.Controls;
+using Avalonia.Controls.Shapes;
 
 namespace FluentAvalonia.UI.Controls
 {
@@ -10,6 +12,18 @@ namespace FluentAvalonia.UI.Controls
     /// </summary>
     public class PathIcon : IconElement
     {
+        private Matrix _transform = Matrix.Identity;
+        private Geometry _renderedGeometry;
+
+        static PathIcon()
+        {
+            StretchProperty.OverrideDefaultValue<PathIcon>(Stretch.Uniform);
+            StretchDirectionProperty.OverrideDefaultValue<PathIcon>(StretchDirection.DownOnly);
+           
+            AffectsMeasure<Shape>(StretchProperty, StretchDirectionProperty, DataProperty);
+            AffectsRender<Shape>(StretchProperty, DataProperty);
+        }
+
         /// <summary>
         /// Defines the <see cref="Data"/> property
         /// </summary>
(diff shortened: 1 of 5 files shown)


amwx:
I think the better solution here would be to add the `Stretch` and `StretchDirection` properties that `Path` has to `PathIcon` and `PathIconSource`. That way the specific scaling behavior then is user choice.

timunie:
I'll try to implement this. It's hard to center the icon though, so I need some more time to do.

timunie:
Ok I think I have found a solution which seems to work the way I need it. Can you please double-check? 

Note: Other then Avalonia does I also center the Icon as I think most of the time icons should be centered. One can still use Aligments or Margin to change this.

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
```diff
@@ -9,6 +12,18 @@ namespace FluentAvalonia.UI.Controls
     /// </summary>
     public class PathIcon : IconElement
     {
+        private Matrix _transform = Matrix.Identity;
+        private Geometry _renderedGeometry;
+
+        static PathIcon()
+        {
+            StretchProperty.OverrideDefaultValue<PathIcon>(Stretch.Uniform);
+            StretchDirectionProperty.OverrideDefaultValue<PathIcon>(StretchDirection.DownOnly);
```
amwx, line 21: Make the default value for StretchDirection `Both`, that way default behavior doesn't change. I'm ok with using `Uniform` as the default stretch - that \*shouldn't\* cause any issues I hope (and now behavior can change anyway)

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
```diff
@@ -25,41 +40,234 @@ public Geometry Data
...
+            Shape.StretchProperty.AddOwner<PathIcon>();
+
+        /// <summary>
+        /// Gets or sets a <see cref="Stretch"/> enumeration value that describes how the shape fills its allocated space.
+        /// </summary>
+        public Stretch Stretch
         {
-            return Data?.Bounds.Size ?? base.MeasureOverride(availableSize);
+            get { return GetValue(StretchProperty); }
+            set { SetValue(StretchProperty, value); }
```
amwx, line 55: nit: change to below for consistency
```C#
get => GetValue(StretchProperty);
set => SetValue(StretchProperty, value);
```

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
```diff
@@ -25,41 +40,234 @@ public Geometry Data
...
+            return (size, transform);
+        }
+
+        public override void Render(DrawingContext context)
+        {
+            var geometry = RenderedGeometry;
+            if (geometry == null)
+                return;
+
+            context.DrawGeometry(Foreground, null, geometry);
```
amwx, line 270: This still needs the clip applied to it as in the old code. Or you can override the default value of `ClipToBounds` (to true) in the static ctor to auto handle this

FluentAvalonia/UI/Controls/IconElement/PathIconSource.cs
```diff
@@ -23,5 +23,35 @@ public Geometry Data
...
+        public static readonly StyledProperty<Stretch> StretchProperty =
+            PathIcon.StretchProperty.AddOwner<PathIcon>();
+
+        /// <summary>
+        /// Gets or sets a <see cref="Stretch"/> enumeration value that describes how the shape fills its allocated space.
+        /// </summary>
+        public Stretch Stretch
+        {
+            get { return GetValue(StretchProperty); }
+            set { SetValue(StretchProperty, value); }
```
amwx, line 39: nit: same as above

amwx (approved):
Just a couple minor things, but looks good to me. Thanks for doing this!

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
timunie: Hi @amwx , 

that was one of those things that made me wonder. If I use `Both`, the icon will look like this if you don't specify a width / height (which is the default): 

image

So I made the icon scale down only be default, so it respects the size of the icon itself. The behavior can be controlled via `<Style Selector="ui|PathIcon" />` but not in code I think.  I've added the needed styles, please double-check if you find this solution OK. If not, please guide me. 

Note: In the original code you have: 
[https://github.com/amwx/FluentAvalonia/blob/fccd2716276d5bf5f22fea0dd8e4b695ba3c2c44/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs#L38-L41](https://github.com/amwx/FluentAvalonia/blob/fccd2716276d5bf5f22fea0dd8e4b695ba3c2c44/FluentAvalonia/UI/Controls/IconElement/PathIcon.cs#L38-L41)

Which will measure the path and therefore also only downscale If I am not missing anything.

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
timunie: done

FluentAvalonia/UI/Controls/IconElement/PathIconSource.cs
timunie: done

FluentAvalonia/UI/Controls/IconElement/PathIcon.cs
timunie: done with overriding in the static constructor, I find this better, as you can control it via style.

amwx:
I'm gonna go ahead and merge this as is. There's a couple things I'll come back to visit later (like the issue where the icon sizes too large when StretchDirection.Both is set without width/height), those'll be added to my checklist for the 11.0 upgrade

timunie:
> I'm gonna go ahead and merge this as is. There's a couple things I'll come back to visit later (like the issue where the icon sizes too large when StretchDirection.Both is set without width/height), those'll be added to my checklist for the 11.0 upgrade

I am very sorry I was not able to do this PR better. But I am only a hobbyist, so some things are beyond my knowledge. 

Anyway, if you think other icons should also add this Stretch properties (SymbolIcon? ImageIcon?) please let me know. I can pick this up if you want me to.

PR #151, opened 2022-05-30
Don't unnecessarily remove page from frame cache

PhantomGamers:
The frame cache is already pruned when a new entry is added that would have caused it to exceed the cache size limit.

Perhaps there was a reason that this was done that I am missing, but I think the current behavior doesn't make sense as if an app has only, say, two pages, and the user goes back and forth between the two pages, the page instances will constantly be remade needlessly.


amwx (approved):
I think my intention was to remove it from the cache when loaded and return it when unloading. Clearly I only implemented half of that idea. Thanks!

Issue #147, opened 2022-05-29
Allow specifying default theme as a fallback


PhantomGamers:
**Is your feature request related to a problem? Please describe.**
Currently when detecting the system theme fails, FluentAvaloniaTheme defaults to the LightModeString.

**Describe the solution you'd like**
Developers should be able to specify the default theme that is used when auto detection fails.

Also just a side note but wouldn't it be more appropriate for the *ModeString variables to be put into an Enum instead?

amwx:
Detecting the system theme shouldn't fail and if it does then something else is wrong. The method I use now uses the WinRT APIs (`Windows.UI.ViewManagement.UISettings` for AccentColor & light/dark mode and `Windows.UI.ViewManagement.AccessibilitySettings` for high contrast mode). Both of these begin with Windows 10.0.10240 (1507) which means all Windows versions I support have these available.

PhantomGamers:
It fails on Linux (and presumably OSX).

amwx:
Resolving the system theme is only supported on Windows at the current time, which is why the property that controls that is `UseSystemThemeOnWindows`. For non-windows platforms, if you don't want to launch in the default light mode, change the `RequestedTheme` property when your app starts.

PR #152, opened 2022-06-01
FATheme add support for getting system theme/accent for Mac/Linux

amwx:
Highly, highly experimental feature here... But, an attempt to allow resolving the system theme and/or accent color for MacOS and Linux.

Changes:
Added two new properties to `FluentAvaloniaTheme`
`PreferSystemTheme` and `PreferUserAccentColor`. Both default to `true`
The existing `UseSystemThemeOnWindows` and `UseUserAccentColorOnWindows` properties are now deprecated and just map to the new properties and will be removed in v2.0. 

While FATheme does not listen for system changes, `InvalidateThemeingFromSystemThemeChange` is now available for mac/linux and will update the theme and accent color accordingly (if the above properties are set to true)

A couple caveats:
1 - The checks for Mac/Linux are quite simple. On Mac, it checks the terminal command `defaults read -g AppleInterfaceStyle` and if that returns "Dark", dark mode is used. On Linux, it requires gtk to work (so things like KDE won't), and queries `gsettings get org.gnome.desktop.interface gtk-theme`. If the return value contains '-dark', dark mode is used. 

2 - Accent color retrieval does not work on linux, as no concept of accent color exists in a standardized way (as far as I'm aware). On MacOS, it queries `defaults read -g AppleAccentColor`. This either returns not found, or an integer 0 - 6. The following assumes MacOS 11 (Big Sur) mappings:
0 - red, 1 - orange, 2 - yellow, 3 - green, 5 - purple, 6 - pink,
Any other values will map to blue. Note that multicolor and graphite settings are not respected and will fall back to blue. Graphite specifically because it's grey, and that won't work well generating the light/dark shades required for the fluent theme.

I only have limited testing capabilities on Linux, and no way to test OSX, so please feel free to test this branch if you have non-Windows available (esp. Mac).
If testing on linux, remember `gtk-theme` must be available (run the above command in the terminal to check).

### DIFF
--- a/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
+++ b/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
@@ -29,6 +29,14 @@ namespace FluentAvalonia.Styling
             _baseUri = ((IUriContext)serviceProvider.GetService(typeof(IUriContext))).BaseUri;
         }
 
+        /// <summary>
+        /// Gets or sets the desired theme mode (Light, Dark, or HighContrast) for the app
+        /// </summary>
+        /// <remarks>
+        /// If <see cref="PreferSystemTheme"/> is set to true, on startup this value will
+        /// be overwritten with the system theme unless the attempt to read from the system
+        /// fails, in which case setting this can provide a fallback. 
+        /// </remarks>
         public string RequestedTheme
         {
             get => _requestedTheme;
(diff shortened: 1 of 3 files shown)


PhantomGamers:
Does this change your opinion of #147 at all? Since it is possible for Linux theme detection to fail if gtk is not present?

amwx:
> Does this change your opinion of #147 at all? Since it is possible for Linux theme detection to fail if gtk is not present?

If the system lookup fails, `RequestedTheme` is given priority as the fallback. If that is unset or invalid, then it falls back to "Light".

PR #155, opened 2022-06-03
Move multitarget from net5.0 to net6.0

amwx:
Changes the targeting from net5.0 to net6.0 and updates the sample and sandbox apps to be net6.0

### DIFF
--- a/FASandbox/App.axaml.cs
+++ b/FASandbox/App.axaml.cs
@@ -2,23 +2,22 @@ using Avalonia;
 using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Markup.Xaml;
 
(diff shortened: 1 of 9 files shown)


PhantomGamers:
Just FWIW as far as github actions goes, with dotnet you can build for all platforms regardless of the platform you're working on, so you don't actually have to build it on linux, windows, and mac for their respective versions.

The one caveat to this is that if you build a gui app for windows on linux (and probably osx), it seems to create a console window in addition to the app gui, so on my project https://github.com/PhantomGamers/SFP I just build all 3 from windows-latest.

Also as far as github actions goes, actions/checkout is now up to v3 and setup-dotnet is up to v2

PR #161, opened 2022-06-07
Improve theme detection on Linux

PhantomGamers:
If the gsettings process exits with exit code 0 we can assume that theme detection is supported on the system and so if the output result does not include -dark we should set the requested theme to Light mode.

Perhaps the exit code check should be added for OSX too, but as I don't have an OSX environment to test it on to know the exit code behavior when dark mode is not set I left it out, and this shouldn't be an issue anyway because as far as I can tell this method of theme detection works on every version of OSX supported by dotnet.

### DIFF
--- a/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
+++ b/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.cs
@@ -603,10 +603,7 @@ namespace FluentAvalonia.Styling
                 var str = p.StandardOutput.ReadToEnd().Trim();
                 p.WaitForExit();
 
-                if (str.Equals("Dark", StringComparison.OrdinalIgnoreCase))
-                {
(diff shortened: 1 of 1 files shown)


PhantomGamers:
Just a note on this but I had one user with a Mac test the theme detection and it didn't work for them, the app starts in light mode even if the system theme is dark mode. They are using an M1 Mac though and my app is compiled against osx-x64, so maybe this is an issue with Rosetta. The `defaults read -g AppleInterfaceStyle` command works as expected for them in terminal, but the FluentAvalonia's check doesn't work, so maybe it is an issue with the Rosetta x86->arm translation layer, I'm not sure.

Ultimately this isn't a huge deal for me as I implemented an option in my app for users to specify the app theme anyway but it's something to keep in mind

amwx (approved):

PR #162, opened 2022-06-07
Some improvements

AigioL:


amwx:
Sorry didn't mean to seemingly ignore this PR. At this point though I'm not interested in taking most of the changes here. I did, however, refactor some of the stuff in `CoreWindow` (in v2-preview2) to be smarter about the platform checks and move platform specific stuff to dedicated, non-inlined methods so hopefully that will fix that issue - following what you did here, I'm just not using the attributes that aren't available in .netstandard2.0.

PR #196, opened 2022-09-13
fix: fix ButtonSpinner sample crush and direction error.

rabbitism:


amwx (approved):

amwx:
Thanks for fixing this.

PR #210, opened 2022-10-14
[v2] Style improvements

### LINKED ISSUE #128: macOS Support
Hi, Everyone
I built this project on macOS 10.15 after I disabled the things related to the Windows Interop.
The building and running are good but the vertical alignment of the text in all the controls is strange.

image

Does not the styling of this support on macOS?
Or how to fix it?

Thank!

### LINKED ISSUE #209: In ComboBox the value is shifted vertically
**Describe the bug**
In ComboBox and ListBox the text is shifted vertically. There is no problem when running the same application on Windows 10/11.

**Screenshots**
Screenshot from 2022-10-13 19-21-47

Screenshot from 2022-10-13 19-27-17

Screenshot from 2022-10-13 19-49-04

 - OS: Ubuntu 22.04.1 LTS
 - FluentAvalonia Version 2.0.0-preview2
 - Avalonia Version 11.0.0-preview2

**Additional context**
Observed even in the [example](https://github.com/amwx/FluentAvalonia/tree/master/FluentAvaloniaSamples).

amwx:
Some updates to the styles here, including a big one that should help on non-Windows platforms:

On account of #128 & #209, I'm adding a way to help adjust text positioning in some controls when not on Windows. Right now we have a problem that Fluent design is primarily targeted at Windows where Segoe UI or Segoe UI Variable font families are available, leading to some controls having text positioned wrong on Mac/Linux where the fonts aren't available. This is done because fluent design says the first line of text should be level with the control (accounting for multi-line scenarios) so vertical alignment is set to top and margins/paddings are used to position the text - based on Segoe UI. The worst offenders of this were `CheckBox`, `RadioButton`, and `ComboBox`. 

To try and fix this, <s>I've added a property to FluentAvaloniaTheme `UseStyleOverridesForNonWindows`, which defaults to true. </s>If set, and not on Windows,</s> some styling overrides are injected to use vertical alignment and adjust some of the uneven margins to make things look a little better. These overrides are added internally inside FATheme, so you can still override further if needed. This isn't a perfect solution <s>because this may lead to different appearances in Windows vs other platforms (primarily if multi-line text is needed),</s> but this is as much as I can do on my end at this point and I think its a fair solution

I've only addressed the 3 controls above. The rest by my eye look the same on Windows or aren't affected. 

[WSL2 ubuntu linux]
Before:
image

After:
image

[Other]
- [Improvement] Change padding of `CheckBox` to use `DynamicResource` instead of `StaticResource`
- [Improvement] Update Padding of `ListBoxItem` to match WinUI (left margin increased from 12 to 16)
- [Cleanup] Removed the old DataGrid axaml file that I forgot to delete
- [Cleanup] Removed some unused resources from the theme resource dictionaries. These were all marked in "Legacy Brushes" section and/or had the "ThemeBrush" suffix in the name. These are unused in WinUI and I don't know why MS still has them in the Fluentv2 styles, but we don't need them so they're gone
- [Fix] Updated the expander style to improve template and fix issue where expander header was rounded on all sides even when expanded

fixes #209 
fixes #128

<b>Edit:</b>
I've changed the API a little. Instead of a bool and only applying on non-Windows, it's now an enum based on @robloo's suggestion below, and the property is now called `TextVerticalAlignmentOverrideBehavior`. The default value is `EnabledNonWindows`.

```C#
public enum TextVerticalAlignmentOverride
{
    Disabled, // never use the overrides
    EnabledNonWindows, // only use when not on Windows
    AlwaysEnabled, // always use the overrides
}
```


robloo:
Nice to see these fixes! My only concern here is for apps that want to ensure the same uniform look on all platforms. In that case they may want to use the override styles on all windows (I would probably use that as the default myself). In that case I wonder if the option shouldn't be renamed to something more specific like `ForceVerticalAlignmentToCenterOfText` and then support three states `Disabled`, `EnabledNonWindows`, `Enabled` (last for all platforms).

robloo:
> These were all marked in "Legacy Brushes" section and/or had the "ThemeBrush" suffix in the name. These are unused in WinUI and I don't know why MS still has them in the Fluentv2 styles

They were kept in for compatibility to not break existing apps so much with the changes for Fluent v2.

amwx:
> Nice to see these changes!
> 
> My only concern here is for apps that want to ensure the same uniform look on all platforms. In that case they may want to use the override styles on all platforms including windows (I would probably use that as the default myself). In that case I wonder if the option shouldn't be renamed to something more specific like `ForceVerticalAlignmentToCenterOfText` and then support three states `Disabled`, `EnabledNonWindows`, `Enabled` (last for all platforms).

I had a similar thought, just thinking of the best approach. I like the enum idea though, might be what I end up doing. I need to think on the property name some though

PR #219, opened 2022-10-23
Improve system theme and accent color detection on Linux

taj-ny:
This PR adds and improves system theme and accent color detection for over 6 Linux desktop environments (KDE, GNOME, Xfce, Cinnamon, LXDE and LXQt). The table below shows how it's supported on each one of them.

|DE|Accent color|System theme|
|----------|-------------|------|
|KDE|Supported (from current scheme, from wallpaper and custom)|Partially supported - checks if the current color scheme name contains ``dark``.|
|GNOME, Xfce, other|Not supported|Supported - checks for the value of ``org.gnome.desktop.interface.color-scheme`` (``default``, ``prefer-light``, ``prefer-dark``). If the value is ``default``, checks if ``org.gnome.desktop.interface.gtk-theme`` contains ``dark``.|
|Cinnamon|Not supported|Partially supported - checks if ``org.cinnamon.desktop.interface.gtk-theme`` contains ``dark``.|
|LXDE|Supported (from custom selection color)|Partially supported - checks if the current theme name contains ``dark``.|
|LXQt|Supported (from selection color)|Partially supported - checks if the current theme name contains ``dark``.|

# Notable changes
- The value of ``org.gnome.desktop.interface.color-scheme`` (``default``, ``prefer-light``, ``prefer-dark``) is now preferred over ``org.gnome.desktop.interface.gtk-theme``. The second one is used only if the first one is ``default``.
- A theme is now considered dark if it contains ``dark``, previously ``-dark``.

# TODO
- [x] Move the new logic into a static helper class


amwx:
Nice, thanks for this! Would it be possible to split the new logic out into a static helper class? FATheme is already pretty crowded and I'd like to avoid the two string fields on FATheme if they only apply to one OS, if at all possible. (I was already thinking about doing something similar with the Windows stuff too)

amwx (approved):
Looks good. Thanks again for this!

PR #220, opened 2022-10-23
Fix IApplicationSplashScreen.RunTasks not being awaited

taj-ny:
This PR fixes the splash screen closing before ``IApplicationSplashScreen.RunTasks`` has finished running.


amwx:
Thanks for catching this! However, since I'm replacing CoreWindow (PR #222) I'm not going to merge this in. I did, however, fix this in the new `AppWindow` and left a comment about this crediting you with the fix.

PR #222, opened 2022-10-28
New: FluentAvalonia.UI.Windowing & AppWindow

### LINKED ISSUE #221: [v2] [Breaking Change] Important changes coming to FluentAvalonia
Over the weekend, it was pointed out that having a reference to `Avalonia.Desktop` can lead to issues (specifically to WASM), at least in the 11.0/v2 world. This reference is necessary for `CoreWindow` as that is a desktop API, which means a solution is necessary to facilitate this in way that works for all, including consumers of FA for non-desktop targets.

Beginning in v2-preview 3, `CoreWindow` will be deprecated and removed from FluentAvalonia. For those that use `CoreWindow`, don't worry, the concept isn't disappearing. A new package `FluentAvalonia.UI.Windowing` will be added to the project to facilitate all Controls/APIs within FA that require the `Avalonia.Desktop` reference. 

**So what's replacing `CoreWindow`?**
A new window class `AppWindow` will be added as a replacement to `CoreWindow`. `AppWindow` will still offer the visual appearance characteristics of its predecessor, however, many of the APIs have been reworked to be simpler. I've managed to fix the known annoying issues that occur on Windows 10 (the missing top window border, the "white outline"). In addition, AppWindow will now properly support dpi scaling, and system theme/accent color changes automatically (Windows only). The default titlebar on Windows will now have an option for an Icon (so you don't need a custom titlebar for that), and extending content into the titlebar/drag rects have been completely overhauled. The old API was built from UWP, this one is similar to `Microsoft.UI.Windowing.AppWindow` (at least in the titlebar perspective). I'll explain more later - initial PR is coming soon.

<s>**TaskDialog**
One side effect of needing to move the windowing APIs out of the main package is that `TaskDialog` will also have to be moved as it has a dependency on `CoreWindow` (soon to be `AppWindow`) when it shows in windowed mode. It will still function the same, you'll just have to include a separate package for it.</s>
EDIT: 
Actually, no I can't do this since that would remove TaskDialog from being used on non-desktop platforms (where it falls back to overlay). I think what I'll end up doing is adding an AppBuilder extension to tack on in program.cs, something like `.UseFAWindowing()` which will register a service that TaskDialog can use to build a windowed dialog or fall back to overlay if that service isn't present. I think that's the cleanest solution.

These are the two controls I know are changing. If there's any others I come across later I'll update this post with the details if it will affect your usage.

amwx:
As described in #221, `CoreWindow` is being removed and replaced with `AppWindow` and is moving to a dedicated package for desktop apps `FluentAvalonia.UI.Windowing`. In addition, with this PR `TaskDialog` will require one extra step to use (for having it hosted in a window) since it relies on AppWindow and that's not in the main project anymore. 

**TaskDialog**
I'll quickly cover this change here so it's not lost below in the long description of AppWindow.
For consumers of `TaskDialog`, if you want to use the window hosted task dialog you'll have to make 1 easy change to make sure TaskDialog can create with an AppWindow. In program.cs as part of the AppBuilder stuff, attach `.UseFAWindowing()`. This extension method is available in the FluentAvalonia.UI.Windowing package. Otherwise, you must always pass `true` to `TaskDialog.ShowAsync(bool showHosted)` to use the overlay layer. ShowAsync will throw an `InvalidOperationException` if you forget these steps.

**Why remove CoreWindow?**
`CoreWindow` was initially a private class used in the Sample app to achieve a modern looking window (title bar height remained when maximized, you could still resize drag from the window shadows despite being a client extended window). By request that was made part of FluentAvalonia and its API grew to support several new features. However, there are several known bugs with how CoreWindow works - primarily because `DwmExtendFrameIntoClientArea` is a horrible mess in Windows 10. I've finally figured out the reason and the best way fix to address them...some of which is caused by me having a better understanding of how it all works. 

Bug # 1, The white border on the window frame on Windows 10. That's a complaint raised several times here that on Windows 10 the frame border is white when the system setting "Show accent color on titlebars and windows" is turned off. And this is because Win32 apps have no concept of light/dark mode, it's ALWAYS light - until you force it. I stumbled across an issue on the Windows Terminal repo where they dealt with the same thing, and the solution was just to always put the window in dark mode, regardless of the app's current theme. Note that this is only the Win32 window, your app content will still respect the theme in FATheme. So, AppWindow automatically does that, which turns out it actually matches modern apps. Bug fixed! Note that you should no longer call `FluentAvaloniaTheme.ForceWin32WindowToTheme` for AppWindow instances - leave them in dark mode.

Bug # 2, the missing top border of the window frame on Windows 10. This one was just my own doing. When handling WM_NCCALCSIZE, the entire top part of the frame is hidden by the client area. For some unknown reason, the DWM team made it so that you can either show no title bar or all the title bar in Win32 - meaning we can't just make the top margin 1 to just show the frame and hide the rest of the title bar. What I was missing was, you then have to offset the top of the client area to make the top border show through again. So, windows 10 now has a margin adjustment automatically to make sure the top frame is visible so I don't have to "hack" a top window border in and possibly have it be the wrong color. Bug fixed!

Here's what `AppWindow` looks like on Win10. Notice the border is consistent on all sides and it's not white!
image

But AppWindow isn't just about fixing those bugs, its about extending the API to make things easier for you and allow you to do more. 

- The default title bar now has built in Icon support - no more custom title bars to achieve this!
- `AppWindow` has better support when display scaling > 100% - something that was lacking in its predecessor as the frame adjustments were all hard coded values - now we use scaling and the OS to tell us - which prevents content from being cutoff on the left/bottom/right of the window in high dpi scenarios.
- `AppWindow` also now has full screen support. Note that the caption buttons are hidden in full screen mode and will not automatically show up by moving your mouse to where they usually are. You are responsible for the mechanism that activates and deactivates full screen mode in your app.
- On Windows, `AppWindow` handles `WM_SETTINGCHANGED` to automatically adjust the accent color and app theme if it's changed in Windows settings and you have `PreferSystemTheme` or `PreferUserAccentColor` in FATheme set.
- AppWindow now has a a property `PlatformFeatures` that allow certain platform specfic behaviors to occur. As of now, these are only for windows, but the opportunity is there for mac/linux too. On Windows 11, you can change the window border color thanks to the newly documented and expanded `DwmSetWindowAttribute` function (unfortunately this isn't available on Windows 10). And there's also a feature to let you use the task bar progressbar. You should be familiar with `ITaskbarList3` before using these settings though.

Example of `AppWindow` on Win11 with a yellow border and the taskbar progressbar in error state at 25%
image

**TitleBars in AppWindow**
Finally, the entire concept of Title bars within AppWindow has changed from CoreWindow. CoreWindow had two separate classes for managing title bars and required you to register a custom title bar control, if desired. This was designed for how it works in UWP. No more! `AppWindowTitleBar` class is a combination more closely resembling `Microsoft.UI.Windowing.AppWindowTitleBar`. From there you can set the custom title bar/caption button colors, height, and whether the content should extend into the title bar area. You can access the `AppWindowTitleBar` using the `TitleBar` property on `AppWindow`. This is always non-null, even on mac/linux, so you don't need platform checks, it will just not have effect outside of Windows.

**HitTesting drag regions**
This has also been completely rethought and is very different. Regardless of whether you extend the content into the title bar, AppWindow retains the default drag rect - meaning that the top part of the window will still be treated as the window drag region, even if you have interactable content there. There are a couple ways to solve this:
1- `AppWindowTitleBar` allows you to set custom drag rectangles via `SetDragRectangles(Rect[] value)`. These override the default drag rect to specify specific areas you want to allow dragging. These Rects must be in Window coordinates and should be updated when the window size changes, content changes affecting the title bar area, etc. Pass null to restore the default drag rect.
2- Option 1 is nice, but may be cumbersome to constantly calculate the regions if window size changes or title bar content is complex. I've added an AttachedProperty to AppWindow `AllowInteractionInTitleBar` that you can attach to controls that makes them exempt from this logic. This makes it easy if, for example, you want a search bar at the top of your window but you want the rest of the top to be draggable. Set the attached property to the Textbox and you'll be able to use it normally. 
```Xaml
<!-- xmlns:wnd="using:FluentAvalonia.UI.Windowing" -->

<TextBox wnd:AppWindow.AllowInteractionInTitleBar="True" />
```

Also noting, that hit testing is now done by simple bounds checks, rather than asking the Renderer to hit test visuals in the scene. However, these new methods do present a problem for controls like `TabView` or `NavigationView` where it's not as easy to set the attached property or calculate drag rectangles. For cases like this, you can change the `AppWindowTitleBar.TitleBarHitTestType` property to `Complex` so that it uses the old logic of asking the renderer for hit testing. It then checks the entire tree from the visual under the pointer to see if any control is focusable before dragging. This enables tabs in the title bar with ease. 

**Custom Title bars**
AppWindow no longer requires you to register a custom title bar, because of the changes above to hit testing. All you have to do is extend the content and manage how you want drag rectangles to work using the options described above. 

**IApplicationSplashScreen**
Yes this still works :)

Couple final notes:
`AppWindowTitleBar` does not currently have something like `LayoutMetricsChanged` event. The height is fully customizable by you (defaults to 32) and the left/right insets only change if you change FlowDirection or scaling. As of now, I don't see a need for that event.

Some restructuring of FluentAvaloniaTheme has been done. `ForceWin32WindowToTheme` no longer crashes if called on non-Windows - so you can safely call it without platform checks.

As I'm currently working on redoing the sample app, its integration at this point with AppWindow in minimal. 

To use `AppWindow`, you'll have to include a separate `StyleInclude` in your apps to load the styles within the `FluentAvalonia.UI.WIndowing` project. You can include it right after FluentAvaloniaTheme:
```Xaml
<sty:FluentAvaloniaTheme />
 <StyleInclude Source="avares://FluentAvalonia.UI.Windowing/Styles/FAWindowingStyles.axaml" />
```

Closes #221


robloo:
Nice! Not just WASM but all the mobile backends should work now that the dependency on Avalonia.Desktop is optional.

Maruhl:
First of all, thank you for your hard work.
I have a few questions about this.

1. Will it be possible to hide the minimize/maximize buttons?
2. Just as an interest: Why don't you use https://www.nuget.org/packages/Microsoft.Windows.CsWin32 for the win32 interops?

amwx:
Alright, everything seems to be working as intended. I also double checked on linux (via WSL) and it all works, so I think we're good to go now. 

Packages have already been pushed, but because FluentAvalonia.UI.Windowing is a new package it might take a minute for it to be available. Also note the new package is FluentAvalonia dot UI and not FluentAvaloniaUI like the main package

amwx:
>     1. Will it be possible to hide the minimize/maximize buttons?

On Windows, yes - use the `ShowAsDialog` property. The only effect it has is to hide the minimize/maximize buttons from the managed titlebar (added for TaskDialog). On other OS's, no b/c it's not a feature of a normal Avalonia window with system titlebars

>     2. Just as an interest: Why don't you use https://www.nuget.org/packages/Microsoft.Windows.CsWin32 for the win32 interops?


CsWin32 is still relatively new and I don't like to jump on new things as soon as they come out. I also like having full control over the pinvoke functions rather than having them auto configured by a couple of files. There's also no real performance gain (since its just a source generator) from this as far as I'm aware so there's no harm in doing it the "old fashioned way"

PR #223, opened 2022-10-28
Preview3 Updates

robloo:
@amwx This gets FluentAvalonia to build against preview3.

- Updated only the necessary nugets
- Removed header from Date/TimePicker to match upstream (following WPF now where there are no headers)
- Updated ColorPicker removing converter usage


amwx:
Thanks. Since this bumps the package references I'll merge this right before I do the AppWindow PR, which should be sometime this week after which I'll release -preview3 here

PR #227, opened 2022-10-31
Add some FlowDirection fixes

danielmayost:
This PR make some small things:

1. Add option to change `MainWindow`'s `FlowDirection` in Setting tab in SampleApp, I accessed MainWindow through `ApplicationLifetime`, if you don't like the way I did it, it can be changed.
2. Make V symbols in CheckBox and MenuToogle as LTR so them appear correctly in RTL mode.
3. Add .vscode directory to the .gitignore, I work with vs code and it is not trivial to delete the files before each PR.


amwx (approved):
Thanks. Looks good enough for me

PR #232, opened 2022-11-06
NavigationView indicator animations for OSX

christosk92:
On MacOS, the NavigationView did not have animations when changing the selected item on the navigation view.

This PR fixes that by implementing a custom behaviour, which animates the SelectedIndicator border (on Left display mode).

Before:

https://user-images.githubusercontent.com/13438702/200153392-b081413a-336e-4ed0-9d6a-da90c89b2cca.mov

After:

https://user-images.githubusercontent.com/13438702/200153394-1a214835-2645-49a3-9ebd-9635198fc94b.mov

### DIFF
--- a/FluentAvalonia/Core/NavigationViewItemSelectedIndicatorState.cs
+++ b/FluentAvalonia/Core/NavigationViewItemSelectedIndicatorState.cs
@@ -1,0 +1,171 @@
+using System;
+using System.Threading;
+using Avalonia;
+using Avalonia.Animation;
+using Avalonia.Animation.Easings;
+using Avalonia.Controls;
+using Avalonia.Layout;
+using Avalonia.Media;
+using Avalonia.Styling;
+using Avalonia.VisualTree;
+
+namespace FluentAvalonia.Core;
+
+internal class NavigationViewItemSelectedIndicatorState : IDisposable
+{
+	private readonly Easing _bckEasing = new SplineEasing(0.2, 1, 0.1, 0.9);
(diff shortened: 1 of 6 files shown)


christosk92:
The video above is a video from my own project. The navigation view now behaves like this in the samples app:

https://user-images.githubusercontent.com/13438702/200153602-5d8aa8a2-68a3-4f07-af01-06f54e9ad656.mov

amwx:
Hi! Thanks for taking an interest in this. Unfortunately, this isn't something I'm looking to merge in. The selection indicator doesn't animate on any OS at the moment. The animations that do it in WinUI are done using composition animations, which we didn't have in Avalonia at the time I ported the control so I left it out (but kept the method where it begins so it could be added in the future - hence the `AnimateSelectionIndicator`). I am looking to add that code in now that we have a compositor, but I'm waiting for it to stabilize a bit and a feature (scoped batch animations) to be implemented first. 

What you've done is pretty neat and may be useful to others looking for this in the near-term, though I don't think its suitable for the library itself. This feature would need to work on all OS's, all display modes of the NavigationView, support hierarchical scenarios, and be properly clipped if the animation has to proceed to an item that's offscreen (scrolling). I also think it'd be better not to be a markup extension and implemented directly in the source and I'd like to avoid setting `ClipToBounds=False` as that could potentially cause issues for others as a result - and this is where composition animations help. Sorry, but thanks again for your interest!

christosk92:
I understand. But what I noticed is that it does work in Windows out of the box! (Windows 11 tested). I will just keep using the code for now in my own app, but if anyone else is interested in using this, they can use my fork however they like.

Cheers!

PR #248, opened 2022-11-25
Add a parameter for ContentDialog.ShowAsync() to specify the target window

TSRBerry:
We are currently running into Linux specific issues with `ContentDialog` at [Ryujinx](https://github.com/Ryujinx/Ryujinx), because for some reason our overlay window still has `IsActive == false` when ShowAsync() is called. On Windows it works fine, but on Linux we couldn't find a way to force `IsActive == true` even tho we were able to make sure it's true before we call ShowAsync().

This is why I created this method. Using this method we get consistent behavior across all platforms and don't need to worry about `IsActive`. Usually we would just use a subclass of `ContentDialog` for this, but since it accesses a lot of private variables, this wasn't really easy to do.

I bet this method also has other usecases, but I figured it might be best to explain why this was needed in the first place.

### DIFF
--- a/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
+++ b/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
@@ -8,7 +8,6 @@ using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Layout;
 using Avalonia.Logging;
-using Avalonia.Styling;
 using Avalonia.Threading;
 using Avalonia.VisualTree;
 using FluentAvalonia.Core;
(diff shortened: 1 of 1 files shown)


FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
```diff
@@ -186,9 +186,67 @@ public async Task<ContentDialogResult> ShowAsync(ContentDialogPlacement placemen
...
+        return await _tcs.Task;
+    }
+
+    /// <summary>
+    /// Shows the content dialog on the specified window asynchronously.
+    /// </summary>
+    /// <remarks>
+    /// Note that the placement parameter is not implemented and only accepts <see cref="ContentDialogPlacement.Popup"/>
+    /// </remarks>
+    public async Task<ContentDialogResult> ShowAsync(Window window, ContentDialogPlacement placement = ContentDialogPlacement.Popup)
```
amwx, line 204: To avoid duplicate code:
Make this private and rename it `ShowAsyncCore` 

Then we can have two public methods to call ShowAsyncCore matching the existing and new API (I'm purposefully ignoring the placement arg since it's technically unused, probably should remove it, but that's technically a breaking change):
`public async Task<ContentDialogResult> ShowAsync() => ShowAsyncCore(null)`
and 
`public async Task<ContentDialogResult> ShowAsync(Window w) => ShowAsyncCore(w)`

TSRBerry:
How does that look?

FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
```diff
@@ -29,6 +28,9 @@ namespace FluentAvalonia.UI.Controls;
 /// </summary>
 public partial class ContentDialog : ContentControl, ICustomKeyboardNavigation
 {
+    public async Task<ContentDialogResult> ShowAsync() => await ShowAsyncCore(null);
```
amwx, line 31: nit: move these below just above ShowAsyncCore for consistency

amwx (approved):
Minor nit, otherwise looks good, thanks!

Also, I'll do a formal deprecation/removal of `ContentDialogPlacement` later in a separate PR so I won't worry here about preserving the optional parameter.

TSRBerry:
Could we also backport this to 1.4.4? So we could use this feature with our current version?
If you are okay with that should I create another PR or would this be handled differently?

amwx:
> Could we also backport this to 1.4.4? So we could use this feature with our current version? If you are okay with that should I create another PR or would this be handled differently?

This \*should\* be possible since its a fairly straight-forward change - but `master`'s source is v2/11.0 so I can't just cherry-pick the commit. I'll see what I can do sometime this week.

amwx:
> I thought the code for 1.4.4 for ContentDialog is the same as for master right now, so cherry-picking that commit should work, but if it's different and I'm misremembering then I'm not sure what the best approach would be.
> 
> Thank you for helping us! :D

#242 is the most recent change to ContentDialog which are v2 fixes only (theres a couple breaking changes and new APIs that rely on stuff added only in the v2 stuff), and there were a couple changes within `ShowAsync` that would've conflicted. I'm pretty much full steam ahead on v2/11.0 now, and because there's so many big changes on going most stuff won't be backported unless its a critical bug fix. 
Anyway, I've pushed `1.4.5` package with this fix in it so hopefully that helps

TSRBerry:
Thank you very much!

PR #249, opened 2022-11-26
Set ContentControlThemeFontFamily for unhandled platforms

Raymonf:
I use this library on mobile for its excellent ContentDialog control, but said control crashes on iOS with "Static resource 'ContentControlThemeFontFamily' not found."

I tracked this down to each call to `Resolve[Platform]SystemSettings()` setting at least `FontFamily.Default` as the default font, but since iOS isn't handled by any of the desktop platform checking branches, the static resource never gets added.


amwx (approved):
Thanks!

PR #253, opened 2022-12-03
[WIP] FluentAvalonia v2-preview5

### LINKED ISSUE #182: TaskDialog Commands property is readonly
**Is your feature request related to a problem? Please describe.**
Only way to set a list of TaskDialogCommands is to create them in xaml template or while initializing TaskDialog from code (although you must provide them statically) and then you can't change them and cannot bind to property to set in template from model.

**Describe the solution you'd like**
Have a bindable property Commands of TaskDialog.

**Describe alternatives you've considered**
```xaml
<ui:TaskDialog>
    <ItemsControl Items="{Binding}">
        <ItemsControl.ItemTemplate>
            <DataTemplate>
                <ui:Button>
                    <StackPanel>
                        <TextBlock Text="{Binding Text}" />
                        <TextBlock Text="{Binding Description}" />
                    </StackPanel>
                </ui:Button>
            </DataTemplate>
        </ItemsControl.ItemTemplate>
    </ItemsControl>

    <ui:TaskDialog.Buttons>
        <ui:TaskDialogButton Text="Cancel" DialogResult="Cancel" />
    </ui:TaskDialog.Buttons>
</ui:TaskDialog>
```

### LINKED ISSUE #261: [TabView] TabViewItem memory leak
### Discussed in https://github.com/amwx/FluentAvalonia/discussions/260

<div type='discussions-op-text'>

<sup>Originally posted by **Haltroy** December 19, 2022</sup>
Removing the tab from the list is just isn't closing it completely, It's still is in memory for some reason. What's the best and proper way to close a tab? </div>

Already know the issue:
https://github.com/amwx/FluentAvalonia/blob/a00da259be01ec153d5b2aa797522cfc5a9f056f/FluentAvalonia/UI/Controls/TabView/TabViewItem.cs#L132-L137

`TabView` is holding a strong reference to each `TabViewItem` from these events which isn't cleared when the TabViewItem is removed from the UI. Probably need to use WeakEvent pattern here or at least make sure the events are properly revoked

### LINKED ISSUE #282: Starting app in Fullscreen mode doesn't work
**Describe the bug**

I am creating an application using the Windowing package that I want to be full screen when it is opened. However when I try to do this, the window will not have any decorations and cannot be dragged as expected. But the window will not expand to take the entire screen space, instead it will be the size as if the WindowState is set to Normal.

I can replicate this issue in my own application, a brand new application made from the avalonia.app template and then had FluentAvalonia installed as well as by modifying the sample application in this repo as well. In all these cases I have added `WindowState="FullScreen"` to MainWindow.axaml.

One mitigation that I have found was to delay making the application full screen until a bit after the application starts. Leaving the WindowState as Normal and then adding this into the window constructor does work correctly.

```cs
Task.Delay(2000).ContinueWith(_ =>
{
    Dispatcher.UIThread.Post(() =>
    {
        WindowState = WindowState.FullScreen;
    });
});
```

Also I did some very limited testing on Ubuntu via WSL 2 and found that the application would start in Fullscreen with no other changes so it seems like a Windows only bug.

 - OS: Windows 11 (10.0.22621.1105) & Windows 10 (10.0.19045.2486)
 - FluentAvalonia Version: 2.0.0-preview4
 - Avalonia Version: 11.0.0-preview4

**Additional context**

I have tested this out on 2 different machines, one with x1.25 scaling and the other with x1 scaling and the issue still occurs.

amwx:
Many changes are coming from upstream, some are already in upstream `master`, so getting a head start here. 

This branch currently builds against `11.0.999-cibuild0028785-beta`.

Because AvaloniaEdit lags behind Avalonia, I'm going to keep it disabled from here until 11.0/v2 is ready to avoid headaches with major changes. Code examples will obviously no longer be enabled in the sample app for the time being.

So far:
- [x] Remove interfaces like IControl/IVisual/etc
- [x] Remove duplicate setters in styles
- [x] Fix CommandBar more button
- [Deferred to future] Move `Frame` animations to Composition APIs instead of "Xaml" animations
- [x] Standardize to IconSource over IconElement (NavigationViewItem, CommandBarButton/ToggleButton, FAMenuFlyoutItem)
- [x] Updates for `TextCommandBarFlyout`
- [x] Fix TabViewItem memory leak (fixes #261)
- [x] Fix ScrollBar corner radius
- [x] Ensure all ContentControls register their ContentPresenter via IContentPresenterHost.RegisterContentPresenter
- [ ongoing ] Various styling fixes/corrections
- [x] Made `TaskDialog` commands and buttons property writeable (closes #182)
  - Note that while these properties are now writable, they are not runtime mutable (no collection change events). You'll have to close and reopen the dialog for changes to take effect. 
- [x] Made the default values for `Stretch` and `StretchDirection` on `PathIconSource` match `FAPathIcon`
- [x] Move some DirectProperties to StyledProperties
- [x] Fix AppWindow starting with `WindowState = FullScreen` (fixes #282)

Note that FA will maintain a System.Reactive reference for this preview. It was removed upstream but because so much of the property system is built on observables and rxui-like stuff, I have code that requires some stuff in System.Reactive, so it stays for now but will be removed before v2 release

<h3>**IMPORTANT** </h3>
Changes resulting from the ItemsControl refactoring:
Three controls (FAMenuFlyout, SettingsExpander, and TabView) previously allowed you to create items for binded items by through a DataTemplate, e.g.,
```Xaml
<DataTemplate>
    <TabViewItem Header="{Binding Content}" />
</DataTemplate>
```
which allows you to bind in other properties more easily. However, with the new ItemsControl API, this isn't possible anymore so some major reworkings of these controls has been done. 

1) FAMenuFlyout. UWP doesn't allow binding on MenuFlyout, but I added it and did my best to preserve it. The way I'm getting around the new limitation is manually generating the items when the flyout loads rather than running it through the normal generation process - this way the FAMenuFlyoutPresenter never sees your view models, but only the generated menu items. This allows me to keep the binding ability working at the trade off of a little extra overhead.

2) SettingsExpander: As of preview 5, SettingsExpander ItemTemplate will only work for the Content property, you will no longer be able to do this:
```Xaml
<DataTemplate>
    <SettingsExpanderItem Content=... IconSource=... />
</DataTemplate>
```
Doing so will result in a SettingsExpanderItem within a SettingsExpanderItem. Instead, you should use the styling system to apply the other properties, similar to how you would in WPF/UWP. Using a control theme and setting ItemContainerTheme should have the same result

```Xaml
<Style Selector="ui|SettingsExpanderItem}">
    <Setter Property="IconSource" Value="{Binding VMIconSource}" />
</DataTemplate>
```

3) TabView: Same goes for TabView. However, per WinUI's sample app, the following is valid (consistent with how it previously worked):
```Xaml
<muxc:TabView.TabItemTemplate>
    <DataTemplate x:DataType="local1:MyData">
        <muxc:TabViewItem Header="{x:Bind DataHeader}" IconSource="{x:Bind DataIconSource}" Content="{x:Bind DataContent}" />
    </DataTemplate>
</muxc:TabView.TabItemTemplate >
```
This is no longer valid in FluentAvalonia.

The reason for this:
In the old version, we had full control over the generation logic of containers, which meant I could check the data template first before generating the default container. In the new logic, which follows the UWP/WPF API, `CreateContainerForItemOverride` has no information about the item the container is for, thus I can't lookup a template since I don't have data type information. I have no idea how that works in WinUI, nothing in the public source code supports that but somehow it \*just works\*. Technically there is a hack in WPF that could be used (used in MenuItem), but we can't use that because `IsItemItsOwnContainerOverride` works slightly differently in Avalonia and is only checked if `item` is a control and not all the time.


adirh3:
I know this is a draft, but in case you're not aware of it the last changes in Avalonia's master made switching themes in FluentAvalonia glitchy, e.g color resources won't change property.
Thanks as always :)

amwx:
> I know this is a draft, but in case you're not aware of it the last changes in Avalonia's master made switching themes in FluentAvalonia glitchy,

Thanks. Was it a change upstream or was it the change I had to make recently to FATheme (#252)?

>  e.g color resources won't change property.
Do you mean as in theme switching or like changing a specific resource via something like `Resources[key] = newValue`?

adirh3:
> > I know this is a draft, but in case you're not aware of it the last changes in Avalonia's master made switching themes in FluentAvalonia glitchy,
> 
> Thanks. Was it a change upstream or was it the change I had to make recently to FATheme (#252)?
> 
> > e.g color resources won't change property.
> 
> Do you mean as in theme switching or like changing a specific resource via something like `Resources[key] = newValue`?

Master works fine with preview 4, so it's probably related to a change in Avalonia.
You can repro it by changing from Dark/Light and back in the samples app.
Seems.to happen only with FluentAvalonia, and only when switching themes, on startup it works fine for both light/dark.

amwx:
Yeah, this was something I had to deal with in #252 but managed to fix. At this point the only difference between FATheme and the `FluentTheme` upstream is that I load the resource dictionaries on demand based on the selected theme and they load all the Styles/ResourceDictionary files on startup (light and dark), unload, and cache them (which I'd like to avoid). I'll dig further this week and see if I can figure out what change upstream is responsible or if its maybe something of my doing

amwx:
@adirh3 I figured out this is related to removing and attaching ResourceDictionaries (which is how I handle theme switching). I've filed this as a [bug upstream](https://github.com/AvaloniaUI/Avalonia/issues/9691) 

I do have another solution if needed, it appears this doesn't break if I don't do remove/add and just do a replace - not sure why though.

adirh3:
> @adirh3 I figured out this is related to removing and attaching ResourceDictionaries (which is how I handle theme switching). I've filed this as a [bug upstream](https://github.com/AvaloniaUI/Avalonia/issues/9691)
> 
> I do have another solution if needed, it appears this doesn't break if I don't do remove/add and just do a replace - not sure why though.

Thanks! Can you share the solution? So I can apply it in my nightly builds

amwx:
> > @adirh3 I figured out this is related to removing and attaching ResourceDictionaries (which is how I handle theme switching). I've filed this as a [bug upstream](https://github.com/AvaloniaUI/Avalonia/issues/9691)
> > I do have another solution if needed, it appears this doesn't break if I don't do remove/add and just do a replace - not sure why though.
> 
> Thanks! Can you share the solution? So I can apply it temporarily in my nightly builds

Sure, the solution is in the most recent commit

adirh3:
Thanks @amwx!

It works perfectly in the samples app, but interestingly in my app there is issue with black text only in Dark mode - 
image

I just recently updated from pre 2.0.0 to this PR's version and it seems to happen to random TextBlocks, any idea what's causing it?

amwx:
Huh, I can't repro that. (I tried in the Sandbox project and in an external project with dll reference) using a normal `Window` and the Foreground was correct.

> Edit: Interestingly, adding this line in a style for `Window` solved the issue -
> 
> ```
> <Setter Property="Foreground" Value="{DynamicResource TextFillColorPrimary}" />
> ```

I'm guessing your custom window has its own styles/control theme? I just checked the normal Window ControlTheme file and that line is present (though using TextFillColorPrimaryBrush instead of the color, but that shouldn't matter). If the Foreground was never set, it defaults to black which is then inherited down the tree. Interesting though that this didn't crop up till now.

adirh3:
> I'm guessing your custom window has its own styles/control theme? I just checked the normal Window ControlTheme file and that line is present
 
Yes you are right, my Window used to have -
```
        <Setter Property="Foreground" Value="{DynamicResource SystemControlForegroundBaseHighBrush}" />
```
Which is based on old Window style from Avalonia. I guess the default Fluent theme still has that resource but FA doesn't.
Sorry for the false alarm, thanks :)

danielmayost:
Don't know what changed, but in this PR, when FlowDirection=RTL mode all the popups (ComboBox etc) appear in the wrong place.
In the master it works well, also in the master of Avalonia.

amwx:
@danielmayost Had a look into this. I can repro this in Avalonia. This also only affects window popups (overlay popups work fine). 

The cause is a difference between setting the FlowDirection on the Window vs a UserControl (or somewhere not at the top level). My sample app sets it on the Window itself, where as Avalonia's control catalog sets it on "MainView". This leads to an issue in the popup positioning code where the transform matrices essentially cancel each other out:
(code is from the PopupPositionerExtensions class in IPopupPositioner.cs)
```C#
public static void ConfigurePosition(ref this PopupPositionerParameters positionerParameters,
    TopLevel topLevel,
    Visual target, PlacementMode placement, Point offset,
    PopupAnchor anchor, PopupGravity gravity,
    PopupPositionerConstraintAdjustment constraintAdjustment, Rect? rect,
    FlowDirection flowDirection)
{
    positionerParameters.Offset = offset;
    positionerParameters.ConstraintAdjustment = constraintAdjustment;
    if (placement == PlacementMode.Pointer)
    {
        // We need a better way for tracking the last pointer position
        var position = topLevel.PointToClient(topLevel.LastPointerPosition ?? default);

        positionerParameters.AnchorRectangle = new Rect(position, new Size(1, 1));
        positionerParameters.Anchor = PopupAnchor.TopLeft;
        positionerParameters.Gravity = PopupGravity.BottomRight;
    }
    else
    {
        if (target == null)
            throw new InvalidOperationException("Placement mode is not Pointer and PlacementTarget is null");
       
        // Problem occurs here
        var matrix = target.TransformToVisual(topLevel);
        if (matrix == null)
        {
            if (target.GetVisualRoot() == null)
                throw new InvalidOperationException("Target control is not attached to the visual tree");
            throw new InvalidOperationException("Target control is not in the same tree as the popup parent");
        }

        var bounds = new Rect(default, target.Bounds.Size);
        var anchorRect = rect ?? bounds;
        positionerParameters.AnchorRectangle = anchorRect.Intersect(bounds).TransformToAABB(matrix.Value);
```

When configure position is called and the FlowDirection is set to RTL on the Window, both `target` and `topLevel` have a RTL flow direction. So when the matrix transform is done, M11 of the matrix is positive 1 and the calculation ends up positioning as if its LTR. When the window is left as LTR, but the popup's target is RTL, this transform works as expected and M11 is negative 1 producing the correct location

danielmayost:
Ahh, it's my fault, at first when I implemented FlowDirection in Avalonia, I made the TopLevel to never MirrorTransform, then I forgot why I did it, and returned it to normal.

Thank you for review this, i will think how to fix it, maybe i return the TopLevel to never MirrorTransform.

PR #271, opened 2023-01-20
Minor Fixes

timunie:
- Missing DynamicResource in Calendar
- Background defined twice in ButtonSpinnerStyles

### DIFF
--- a/FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarStyles.axaml
+++ b/FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarStyles.axaml
@@ -392,7 +392,7 @@
                                     <ContentPresenter Name="Text"
                                                       Foreground="{TemplateBinding Foreground}"
                                                       Background="{TemplateBinding Background}"
-                                                      BorderBrush="{DynamicResource }"
+                                                      BorderBrush="{DynamicResource CalendarViewBorderBrush}"
                                                       BorderThickness="{TemplateBinding BorderThickness}"
                                                       Content="{TemplateBinding Content}"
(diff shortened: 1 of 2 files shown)


FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarStyles.axaml
```diff
@@ -392,7 +392,7 @@
                                     <ContentPresenter Name="Text"
                                                       Foreground="{TemplateBinding Foreground}"
                                                       Background="{TemplateBinding Background}"
-                                                      BorderBrush="{DynamicResource }"
+                                                      BorderBrush="{DynamicResource CalendarViewBorderBrush}"
```
amwx, line 395: This should be `CalendarViewNavigationButtonBorderBrush` per the [21h1 styles](https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/CommonStyles/CalendarView_themeresources_21h1.xaml). Probably can put that up in a Setter and TemplateBinding it in.

amwx (changes requested):
Small thing in the Calendar fix, but otherwise, thanks for catching/fixing these!

FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarStyles.axaml
timunie: > Probably can put that up in a Setter and TemplateBinding it in.

I thought the same, but wasn't sure if you'd like it. Pushed the requested changes.

amwx (approved):

PR #284, opened 2023-02-04
FluentAvalonia preview 5

amwx:
I stupidly merged the ThemeDictionary PR into master instead of the preview 5 branch (which took the preview5 branch with it), but its fine. This PR will be the last few changes I want to get into preview 5 before releasing it. I'm aiming to have it ready no later than Monday.

I'm mostly focusing on some styling fixes now. 
- Fixed NumericUpDown control theme to behave more like NumberBox
- A couple SettingsExpander related fixes
- Added the styles for RefreshContainer/RefreshVisualizer
- Added a TextCommandBarFlyout as the ContextFlyout for SelectableTextBlock

<h2>NOTE</h2>
The default values of `PreferSystemTheme` and `PreferUserAccentColor` are now false. Technically, they were broken a while ago because FATheme is loaded in the constructor before those properties are set and if you set it to false, it wasn't respected. This really came into play with the new theme stuff because if it loads with true but should be false, the system theme was written out to App.RequestedThemeVariant, which would overwrite your preferred setting and couldn't be reversed. This more affected PreferSystemTheme than accent color, but to be safe I've also switched that to false by default


adirh3:
In case you are not aware of this, seems like all high contrast resources fail to add using `TryAddResource` causing these exceptions to be thrown on startup - 
image
Method `TryLoadHighContrastThemeColors`

amwx:
> In case you are not aware of this, seems like all high contrast resources fail to add using `TryAddResource` causing these exceptions to be thrown on startup - image Method `TryLoadHighContrastThemeColors`

The out of range exception should now be fixed. However, note that the high contrast theme is broken in preview5 b/c of an issue with DynamicResource so most of it probably won't load anyway. That fix wasn't prioritized into preview5 upstream so it has to wait until preview6.

PR #285, opened 2023-02-05
Feature: Enable trimming and NativeAOT support

hez2010:
- Use compiled bindings everywhere, and fix invalid bindings found by the compiler
- Fix trimming issue against json deserialization

Should be merged before #284 get merged because I'm merging into the branch [v2_preview5_final](https://github.com/amwx/FluentAvalonia/tree/v2_preview5_final).

Prebuilt NativeAOT FluentAvaloniaSamples app using nightly Avalonia for testing: [FluentAvaloniaSamples.zip](https://github.com/amwx/FluentAvalonia/files/10611104/FluentAvaloniaSamples.zip)


FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
```diff
@@ -0,0 +1,16 @@
+﻿<Styles
+    xmlns="https://github.com/avaloniaui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    x:Class="FluentAvalonia.Styling.FluentAvaloniaTheme">
+    <Styles.Resources>
+        <ResourceDictionary>
+            <FontFamily x:Key="SymbolThemeFontFamily">/Fonts/#Symbols</FontFamily>
+            <Color x:Key="SystemAccentColor">#00000000</Color>
```
hez2010, line 8: A placeholder, SystemAccentColor will be override at the runtime.

amwx (changes requested):
I would really rather FATheme not have a xaml file. I've purposefully made it code-only so I can control the order in which things are loaded. Other than that, I think everything else is ok.

hez2010:
> I would really rather FATheme not have a xaml file.

It's not possible to make it trimming compatible without a xaml file. Basically the axaml compiler will look for the AvaloniaXamlLoader.Load(this) and replace the line with generated code to make it get referenced at the compile time.

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
```diff
@@ -0,0 +1,16 @@
+﻿<Styles
+    xmlns="https://github.com/avaloniaui"
```
amwx, line 2: nit: xmlns on the same lines as styles, then align the others with that

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
amwx: I can't repro that exception so let's take this out. It's odd that's the one it was tripping on though as most of fluent v2 uses one of the light or dark variants rather than the straight accent color

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    x:Class="FluentAvalonia.Styling.FluentAvaloniaTheme">
+    <Styles.Resources>
+        <ResourceDictionary>
+            <FontFamily x:Key="SymbolThemeFontFamily">/Fonts/#Symbols</FontFamily>
+            <Color x:Key="SystemAccentColor">#00000000</Color>
+            <ResourceDictionary.MergedDictionaries>
+                <MergeResourceInclude Source="/Styling/StylesV2/Fluentv2.axaml" />
```
amwx, line 10: Weird designer error
image

Also if you're going to switch to MergeResourceInclude, you need to go through all of FATheme and make sure any place that was expecting a merged dictionary is corrected. Though IMO we're not gaining that much with MergeResourceInclude here since it's just 1 xaml file that has to load and Fluentv2.axaml is basically already a merged xaml file, unlike upstream where their fluent theme is broken into like 5 different ResourceDictionaries.

amwx (commented):
> The XAML is loaded at the time AvaloniaXamlLoader.Load(this) is called, so you can still determine the order of things to load.

Sort of. Ideally, I really want `ResolveThemeAndInitializeSystemResources` to run first before loading anything else. However, from testing this, no resource lookups are done between `AvaloniaXamlLoader.Load(this)` and that method so this behavior is still maintained I guess. 

I still don't like it being a xaml file but I guess my hands are tied here, so just address the couple of things below and I'll merge it.

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
hez2010: Done.

FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml
hez2010: Done.

amwx (approved):

danielmayost:
@hez2010 How did you deploy FluentAvaloniaSamples? something special? Because for me the exe file comes out around 50mb, for you it came out 35mb.

hez2010:
There have been a bunch of improvements of the NativeAOT compiler, you can use the nightly version of ILC and try again.

To do so, 

1. Add the net8.0 nuget source `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json`
2. Add `<PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-*" />` to FluentAvaloniaSamples.csproj
3. Publish with NativeAOT again `dotnet publish -r win10-x64 -c Release -p:PublishAot=true`

PR #298, opened 2023-02-10
Turn dialog animations back on

amwx:
PR turns the open and close animations for `TaskDialog` and `ContentDialog` back on. I can't repro the original animation issues when the composition renderer was first implemented so I think they've been fixed, though not sure where/how.


danielmayost:
Maybe it changed again somewhere, but I can reproduce the problem again (preview 6).

PR #312, opened 2023-02-25
Next preview/iteration

amwx:
Branch is on Avalonia nightly `11.0.999-cibuild0030643-beta`

I have also pushed a new preview package `preview5.1`, which is currently targets this nightly package, and contains any fixes done here since preview 5 which includes:

- Return of the FAComboBox (#292)
- Avalonia's PathIcon has a ControlTheme now (#297)
- Dialog animations have been turned back on (#298)
- Project was restructured to be a little better organized, and central package management was switched to (#299)
- AppWindow returned to main package (#307)
- A bunch of other fixes in #309

<h2>NOTE TO APPWINDOW USERS</h2>
Beginning with this preview-preview package, per #307, `FluentAvalonia.UI.Windowing` package is now deprecated. AppWindow is now entirely in the main repo, please see #307 for more on this, including changes you'll have to make (mostly just deleting the references to the package - the namespace still exists)


amwx:
> @amwx thanks a lot for keeping up with Avalonia's nightly! I have just updated to it, and seems like the issue where black text is showing in the Dark theme has returned - image
> 
> You can repro that in the FluentAvalonia Samples by going into the Expander page and hovering with mouse on the header of the expanders.

For Expander, it seems like there's some referenced resources with the wrong key. I've fixed the header pointer over thing. I still have a few other things to address here and I need to double check with WinUI's template too so I haven't pushed anything out yet. Was Expander the only thing affected? If there are, or you find others, just list them here and I'll double check the resource keys.

adirh3:
@amwx not only expander, for some reason even `TextBlock`s inside Canvas have this issue.
In the sample app I saw something similar in the `SplitView` control, the content inside has wrong text color as well.
Also, I just checked if this happen when switching to the normal Avalonia theme, and seems like it has similar issue for other controls, like `TextBlock`s inside a `ListBox`; While it's not the exact same maybe the core issue is somewhat in Avalonia.

amwx:
Adding to the mystery here, in the new sample app that's in the works none of these issues are present. Well, the wrong Expander resource keys are, but the side effects aren't there and the TextBlocks in the SplitView example are displaying correctly, so at least in terms of my sample app, I wonder if there's some older code or something that is causing some styling issues? Though, if you can repro it in the Fluent v1 theme, then there may still be some issues upstream.

image

robloo:
The Expander resources didn't really change between Fluent v1 and v2 I thought. Expander only ever really had the v2 designs even in WinUI. If it's a problem upstream I can fix it though.

Separately, I have seen a lot of examples of strange resource lookup behavior and even missing resources. I think there are some deeper issues here after all the styling and theme changes upstream. I've filed various specific issues as they are found but I doubt it covers everything yet.

The text color has been an ongoing issue that pops up once in a while ever since ControlTheme was first merged in.

amwx:
> The Expander resources didn't really change between Fluent v1 and v2 I thought. Expander only ever really had the v2 designs even in WinUI. If it's a problem upstream I can fix it though.

I don't really know, nor how long its been like this, but the resources in the dictionary all have 'Header' in them (e.g., ExpanderHeaderBorderBrush) while in the template/styles that was omitted (just ExpanderBorderBrush). That's been fixed now though.

> Separately, I have seen a lot of examples of strange resource lookup behavior and even missing resources. I think there are some deeper issues here after all the styling and theme changes upstream. I've filed various specific issues as they are found but I doubt it covers everything yet.
> 
> The text color has been an ongoing issue that pops up once in a while ever since ControlTheme was first merged in.

Yes, and it seems these tend to happen in hard to repro situations which is making this even harder to find the cause

robloo:
> I don't really know, nor how long its been like this, but the resources in the dictionary all have 'Header' in them (e.g., ExpanderHeaderBorderBrush) while in the template/styles that was omitted (just ExpanderBorderBrush). That's been fixed now though.

Oh, there was a big refactoring in WinUI I did that separated header from content more clearly. That was before the control was officially released. It reworked the resources too.

Avalonia was still using older names (from unreleased initial WinUI version) rather than the released ones. Then I more recently applied that renaming to Avalonia in October 2022 in PR9270. So Avalonia now marches WinUI resources where it made sense (header/content clearly and logically separated).

When Fluent Avalonia was updated I didn't think to mention this so the old resources were probably not changed but then the new style was copied over from October. So that where the mismatch comes from.

amwx:
Update March 8:
CI build version updated to 31748 & preview5.2 package released

Changes in 5.2:
- Fixes wrong Expander header resource names
- Removed the `Setter` for `ItemsPanel` in the Avalonia ComboBox (which was forcing a normal StackPanel for some reason I don't remember)
- Removed the forced MinWidth in `CheckBox`, `RadioButton`, and `ToggleSwitch`. This is a recurring thing that's brought up and I've decided it's not worth it anymore. To compensate, I've adjusted the padding to make the right padding match the left padding (8px) so there's still a little gap to the right of the content
- Various fixes to address the changes from upstream (mostly FlyoutPopupBase and Direct->Styled property changes)

Next preview will be after the ItemsSource PR stuff is merged upstream. It will either be 5.3 or 6, depending on what upstream does

mysteryx93:
How did you get `31748` ?? The latest available nightly build is `11.0.999-cibuild0031698-beta`

mysteryx93:
ok updated builds are available now; not sure what happened before

amwx:
Branch is now on 31862. A couple of big things here:

1- I'm aware of 2 issues with the NavigationView, which seem to stem from the changes to SplitView upstream. The first, has a workaround now implemented and is a result of the pane events being changed to RoutedEvents. This affects nested NavigationViews (like on the sample page, so its more an edge case, but still something), where the pane events of one of the sample NavViews was trigger pane events on the main app NavView. Added a check to only handle the SplitView events if `e.Source` is the current NavigationView's SplitView.
The second is more complicated an only affects NavigationViews where the SplitView pane is in an "overlay" state - so NavigationView PaneDisplayMode = LeftCompact or Minimal. You won't be able to close the pane if its opened. This I think is some weirdness around `SetCurrentValue`, an issue has been filed upstream.


- `System.Reactive` reference has been fully removed and replaced with a couple internal helpers where the Avalonia code expects observables. However, this required changes in the NavigationView's SelectionModel.

I have rewritten the SelectionModel to be a direct port of WinUI's version, rather than the modified version that currently exists - which was the original Avalonia port done with the ItemsRepeater before it was removed and replaced. The main problem is that the ChildrenRequested event args expected an observable instead of just being an `object` like in WinUI, which led to a minor headache when originally porting the NavView. With System.Reactive gone, I don't have a way to go from IEnumerable -> observable easily, and I already wanted to fix this so, now its fixed.

Anyway, I did some minimal tests on the sample page and everything seems to be working as before - but if you find issues, please open an issue. 

(And before someone asks, if you're wondering why I don't just move to the new SelectionModel in Avalonia - NavigationView's code base is very large and very fragile, so its a don't touch it unless absolutely necessary thing)

amwx:
Branch is now on CI 31947, changes here are mostly around Items -> ItemsSource

- `SettingsExpander`, the overflow presenter for `CommandBar`, `TaskDialog` and `FAMenuFlyout` were updated for the ItemsSource change. These require no changes on your part, mostly internal stuff - with the exception of SettingsExpander as that's an ItemsControl already where you should switch to ItemsSource for binding or providing a list of items
  - *NOTE*: `FAMenuFlyout` will retain only a single `Items` property because of complications that already exist around FAMenuFlyout

*NavigationView*
- Was updated to add the missing `MenuItemsSource` and `FooterMenuItemsSource` properties, aligning the NavigationView with the new Items vs. ItemsSource. This is a forced change and ttempting to set `MenuItems` or `FooterMenuItems` will result in an error now as its a readonly property. 
- Note however, `MenuItems`/`FooterMenuItems` are not like `Items` vs `ItemsSource` in that Items acts as a view over the ItemsSource like in ItemsControl - its either Items or ItemsSource, as in WinUI. Therefore, if you supply [..]ItemsSource, you should only access items via [..]ItemsSource as Items will be empty.

*TabView*
- TabView will continue to be a single `Items` property. Part of TabView's code makes zero sense in what its trying to do, which was never a problem before as we didn't have ItemsSource at the time. This may change in the future, but for now, its staying as is.


Other fixes:
- Removes the layout invalidate hack from `ContentDialog` and `TaskDialog` as it was causing problems (#315). There was a reason for this hack, but I can't seem to repro that anymore so we might be safe?
- Updated `SettingsExpander` to left-align the header/description text if there is no IconSource supplied. This applies to the entire SettingsExpander.
  - Setting IconSource on the main header item, all sub-items, if present, will be indented
  - Setting IconSource on one of the sub-items, all items, including the header item, will be indented
  - However, there is 1 caveat. If the main header item doesn't have an Icon but one of the sub-items does, if the SettingsExpander is collapsed, the sub items haven't materialized yet and therefore the icon placeholder indent won't apply until you expand the SettingsExpander. I'm considering this rather minor as I expect most cases will probably have an icon in the main header, or won't use icons at all. Will monitor feedback though.


This will generally be the last of the content for preview6, which I imagine will come (from upstream) fairly soon. The sample app was only updated for these changes where absolutely necessary. It's replacement is coming very soon after preview6, so I'm mostly leaving it as is

amwx:
Preview 6 has been pushed out. Last couple changes to note:
- Frame now subscribes to `TopLevel.BackRequested` and will handle navigating back if `IsNavigationStackEnabled` and `CanGoBack` are true
- In FluentAvaloniaTheme, the accent colors are now loaded into a dictionary and if swapped at runtime, the entire dictionary is replaced. Helps perf, as now only 1 resource change notification happens instead of 7.

PR #318, opened 2023-03-14
Feature Add Package info

workgroupengineering:
- Add /Assets/FaIcon.png from https://github.com/amwx/FluentAvaloniaDocs/blob/main/Assets/FAIcon.ico
- Add /src/Directory.build.pros for centralizzing package info (Version,Icon,tags)
- Add PackageIcon
- Add support at SourceLink


amwx (approved):
This is fine, I think. 
Two things:
- (this really doesn't matter it'll get overwritten later anyway) but you can bump the package version to preview6 since we're on that now - at least the history won't look like we went back in time 😂
- Can you comment out the stuff that actually includes the icon for the package? It's getting an update soon as part of v2 and I'll add it all back in later for the v2 release.

PR #328, opened 2023-03-28
fix: test compilation issues

workgroupengineering:
align tests to Avalonia Preview 6. Next step, fixing failed tests.


amwx:
Thanks for taking an interest in this, however, I've been in the process of re-writing the unit tests to use the headless platform (which is what's recommended now) but I keep hitting road blocks with things from upstream. Current issue is [this](https://github.com/AvaloniaUI/Avalonia/issues/10686) where the compositor deadlocks on dispose preventing the unit test app from exiting. No idea when that will get fixed. Best thing I can say right now, just unload the project so it hides the errors. Its what I've been doing.

PR #329, opened 2023-03-29
Symbol font update

### LINKED ISSUE #321: Planned changes to Symbol enum & font
Some updates/fixes coming to the `Symbol` enum and font. Primary goal here is to cut down on repetitive icons and icons without glyphs. 

- Several icons have incorrect glyphs, will be corrected
- Symbols in the enum:
  - `Favorite` will removed. `Star` serves the same purpose
  - `CodeFilled` will be removed
  - Add `CodeHTML` value, code will map to braces ('{ }') and CodeHTML will map to ('</>')
  - `MoreHorizontal` will be removed in favor of just `More`. 
  - `Left`, `Down`, and `Right` will be added (`Up` already exists), these icons are arrows pointing in their respective direction
  - `Setting` will be removed, use `Settings` (singular vs. plural)

- Symbols marked as `Obsolete`
Several symbols are marked as Obsolete as they didn't have glyphs from Fluent UI System icons

image

Symbols in green will be added to the font & have `Obsolete` removed. Symbols in yellow and grey still don't exist, but could probably be made manually in inkscape (or any vector editing app, mostly reusing the existing icons), need to figure out if the system that's used to build the font would support this. Note that this is only an option at this point. `Placeholder` will be removed

- Various `Filled` variants of icons will be removed. These all pertain to icons that are mostly lines, like Chevrons, arrows, etc. These aren't truly filled anyway. The normal versions will remain, though

image

I believe most of these are placed in a special area in the font specifically for FluentAvalonia. If so, they will likely be removed from the font too.

Within the actual font, various icons that didn't previously have glyphs, now do and will be added to the font
image

amwx:
Updates the embedded symbol font & `Symbol` enum with a few changes. Actual changes slighly differ from #321

In the `Symbol` enum:
- Added `CodeHTML` symbol - `Code` is two braces ' { } ', while `CodeHTML` is ' </> '
- `MoreHorizontal` has been removed, just use `More`.
- `Admin`, `BrowsePhotos`, `LeaveChat`, `Manage`, `OtherUser`, `ProtectedDocument`, `Remote`, `RepeatOne`, and `Shuffle` have been un-obsoleted as glyphs now exist for them
- The following `Filled` versions of these Icons have been removed, both from the Symbol enum and the font itself:
(The regular/un-filled variants remain)

image

Various icons within the font itself have been corrected or added - those can be found in #321 

This is the last big thing before I can get the new sample app into a release state. The new app will have a better experience for browsing BOTH the icons in symbol enum and the font itself

closes #321


src/FluentAvalonia/UI/Controls/IconElement/Symbol.cs
```diff
@@ -116,7 +100,7 @@ public enum Symbol
     CloudSyncCompleteFilled = 0xF8044,
     CloudSyncFilled = 0xF8045,
     Code = 0xE943,
-    CodeFilled = 0xF8047,
+    CodeHTML = 0xF8134,
```
robloo, line 103: I know you already merged and this is a nit.

But usually it would be `CodeHtml` following standard/modern .net naming conventions. Just letting you know as this is API related and may not be changeable easily in the future.

src/FluentAvalonia/UI/Controls/IconElement/Symbol.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
+﻿// WARNING - DO NOT EDIT THIS FILE MANUALLY
```
robloo, line 1: I'm probably going to add an `/// <auto-generated/>` tag upstream.

PR #331, opened 2023-04-01
Update ColorPicker Control Themes

robloo:
Syncs all ColorPicker control themes with upstream changes.

~This is currently draft because it requires latest Avalonia nightly to build.~


robloo:
@amwx Due to upstream preview 7 release I marked this as ready for review. It can be merged after https://github.com/amwx/FluentAvalonia/pull/345.

Let me know if you want me to rebase it on PR 345 directly though.

amwx:
Apologies @robloo. I forgot about this and have already pushed out preview7. It doesn't look like anything critical was changed as it still seems to function, but if there is let me know and I can push out a 7.1 with this PR.

robloo:
I don't need the changes in this preview and those that do  are using custom styles already so they just needed it in upstream Avalonia. This can wait for RC I think.

PR #337, opened 2023-04-11
Enable CodeViewer in samples

hez2010:
- Enable CodeViewer in samples
- Use Json source generator to avoid trimming issue
  - Change `init` to `set` as json source generator in .NET 7 doesn't support `init` properties yet, we can revert them back once we upgrade to .NET 8

image


robloo:
I know a new Samples app was being worked on. That might make this obsolete.

hez2010:
Closing as a new sample app is coming.

PR #338, opened 2023-04-12
Unsubscribe from ColorValuesChanged on window close

RobertBeekman:
I noticed an issue where any closed window would stay in memory indefinitely.

This is troublesome for applications with tray icons because it means no part of the UI ever gets GC'd after the main window closes.

dotMemory pointed out this event handler as the culprit so I made it unsubscribe when the related window has closed.


amwx (approved):
Darn event handlers :) Thanks!

Issue #262, opened 2022-12-27
Soften Button Border Brushes


robloo:
**Describe the bug**

Fluent v2 uses Acrylic Brushes for various borders -- including Button. Avalonia does not currently support Acrylic. Therefore, my assumption is FluentAvalonia uses the fallback values.

This results in button borders that are "too sharp" in dark theme (likely light as well).

To fix this, I would like to pre-blend the button border brush with the background brush taking into account the acrylic transparency as if it was there. This pre-blended color value would be the new border brush value. This means it would differ from WinUI numerically but look more visually similar (softening the button edges from a design standpoint). 

**Screenshots**

WinUI (using UWP, WinUI 2 Library, Windows 10 Pro):

![image](https://user-images.githubusercontent.com/17993847/209735029-bc7a1bae-57bb-455a-975b-cd32cd5f4c2c.png)

FluentAvalonia:

![image](https://user-images.githubusercontent.com/17993847/209734992-4c018fe8-e4be-4ad6-a5e6-ffa70f93ca2e.png)

**Desktop/Platform (please complete the following information):**
 - OS: [e.g. Windows (11/10), linux (distro), osx, wasm, etc]
 - FluentAvalonia Version [e.g. 1.0.0, or latest]
 - Avalonia Version [e.g. 0.10.10]

**Additional context**

I will fix this once direction is agreed.

amwx:
Actually, Buttons don't use an acrylic brush. IIRC in Fluent v2, Acrylic brushes are only used in overlays (FlyoutPresenter, e.g.).

`ButtonBorderBrush` maps to `ControlElevationBorderBrush` which is a LinearGradientBrush composed of `ControlStrokeColorSecondary` and `ControlStrokeColorDefault` colors

I think the actual issue here is the use of `BackgroundSizing=InnerBorderEdge` in WinUI, which isn't available in Avalonia. A quick test modifying the template to place the ContentPresenter in `FABorder` with the background sizing set results in this, which looks much better.
![image](https://user-images.githubusercontent.com/40413319/209873824-bc08760c-a080-4520-b948-954f1abfc70b.png)

robloo:
Yea, sorry, I should have looked at the XAML and instead just made a guess. What you found makes sense though and the TestButton looks perfect. The good news is this is a lot "cleaner" to fix than customizing a bunch of colors.

[...]

PR #345, opened 2023-04-22
[WIP] preview7

### LINKED ISSUE #330: Separator not showing since 2.0
Hello,

I've been porting over my application from 1.x to 2.0.0-preview6 with great success, thank you for that :)
One issue I've noticed so far is that the `<Separator>` element isn't rendering. I first thought it was some old styling I had but I also noticed it missing in the samples application.

I was also able to reproduce this in a simple new application with the following references
```xaml
    <PackageReference Include="Avalonia" Version="11.0.0-preview6" />
    <PackageReference Include="Avalonia.Desktop" Version="11.0.0-preview6" />
    <PackageReference Include="Avalonia.Themes.Fluent" Version="11.0.0-preview6" />
    <!--Condition below is needed to remove Avalonia.Diagnostics package from build output in Release configuration.-->
    <PackageReference Condition="'$(Configuration)' == 'Debug'" Include="Avalonia.Diagnostics" Version="11.0.0-preview6" />
    <PackageReference Include="FluentAvaloniaUI" Version="2.0.0-preview6" />
```

**Screenshots**
With FluentAvalonia
image
Without FluentAvalonia
image

I went back through all the preview versions of Avalonia and FluentAvalonia, they all show the same behaviour so it must've gone awry somewhere between 1.4.1 and 2.x

Hopefully that's enough information to track the issue down, I went through the code but couldn't really find any styling on the Separator at all, so wasn't able to submit a PR :(

### LINKED ISSUE #333: Cannot compile WASM since preview 5 (Not sure on version)
Since version preview 5 I cannot compile WASM anymore. I get this error:

The return type 'FluentAvalonia.Interop.Win32.LRESULT' of pinvoke callback method 'FluentAvalonia.Interop.Win32.LRESULT WndProcStatic(FluentAvalonia.Interop.Win32.HWND, System.UInt32, FluentAvalonia.Interop.Win32.WPARAM, FluentAvalonia.Interop.Win32.LPARAM)' needs to be blittable.  AvaloniaApplication1.Browser  C:\Program Files\dotnet\packs\Microsoft.NET.Runtime.WebAssembly.Sdk\7.0.4\Sdk\WasmApp.Native.targets  290  

This just "happened" after the upgrade to FluentAvalonia and Avalonia. I cannot even compile the Avalonia Greetings template anymore.

amwx:
WIP towards either preview7 or rc1, whichever upstream decides on.

Branch is on: 11.0.999-cibuild0033637-beta

**Required updates:**
- Items->ItemsSource on remaining Avalonia controls that were updated
- Updated ScrollViewer template based on changed logic
- Dealt with reduced Avalonia public API...which is making me absolutely lose my mind...

Note for users of FAComboBox: 
- Because of the loss of IWindowImpl APIs, using the FAComboBox with `IsEditable=true` where the popup has to flip to the top of the ComboBox (not enough space below), the popup will retain the default below display for rounded corners - the top will be unrounded and the bottom will be. This only affects windowed popups - overlay popups, if enabled, are not affected.

**Changes & Fixes:**
- Added the missing `Separator` `ControlTheme` (fixes #330)
- Added a hack fix to prevent the flickering when opening a ContentDialog or TaskDialog - issue stems from something about the animation timer is de-sync'd from rendering, and who knows if that'll get fixed soon
- Added the logic in the NavigationView for `Loaded` events, which wasn't previously implemented as we didn't have `Loaded` at the time NavView was ported
- Disabled the auto ToolTip in TabViewItem...this needs some thought. Currently the auto tooltip was just using the Content, which was loading the Tab content into the tooltip b/c of the implicit DataTemplate lookup Avalonia does - which isn't ideal. 
- Changes the function pointer signature on net5+ for AppWindow's Win32WindowManager so that it doesn't kill the WASM build (fixes #333)

**Special change:**
Added `FAContentPresenter` control to support `BackgroundSizing`. To address some rendering issues, see #262, I've added this special control to support BackgroundSizing for some controls. Currently this is only implemented on `Button`, `ToggleButton`, and `RepeatButton` where this change makes a noticeable difference. I've mapped the `IStyleable.StyleKey` to the normal `ContentPresenter` meaning if you're using custom styles you do NOT have to make changes here and if in the future BackgroundSizing is added upstream, you don't have to make more changes to undo this...so that's a plus.

Before and After:
image

I've also changed the root border in `Expander` to use a FABorder. This, however, will require updates if you're using custom styles for Expander as FABorder is not mapped to a normal Border. The change ensures the toggle button in the expander header gets the same treatment, and aligns it with SettingsExpander too.

### DIFF
--- a/src/FluentAvalonia/UI/Controls/MenuFlyout/MenuFlyoutInteractionHandler.cs
+++ b/src/FluentAvalonia/UI/Controls/MenuFlyout/MenuFlyoutInteractionHandler.cs
@@ -70,9 +70,10 @@ internal class MenuFlyoutInteractionHandler : IMenuInteractionHandler
             window.Deactivated += WindowDeactivated;
         }
 
-        if (_root is TopLevel tl && tl.PlatformImpl != null)
-            tl.PlatformImpl.LostFocus += TopLevelLostPlatformFocus;
-
+        // v2p6.1: Don't have access to this anymore...
+        //if (_root is TopLevel tl && tl.PlatformImpl != null)
+        //    tl.PlatformImpl.LostFocus += TopLevelLostPlatformFocus;
+        
         _inputManagerSubscription = InputManager?.Process.Subscribe(RawInput);
     }
 
@@ -103,8 +104,8 @@ internal class MenuFlyoutInteractionHandler : IMenuInteractionHandler
             root.Deactivated -= WindowDeactivated;
         }
 
-        if (_root is TopLevel tl && tl.PlatformImpl != null)
-            tl.PlatformImpl.LostFocus -= TopLevelLostPlatformFocus;
+        //if (_root is TopLevel tl && tl.PlatformImpl != null)
+        //    tl.PlatformImpl.LostFocus -= TopLevelLostPlatformFocus;
 
         _inputManagerSubscription?.Dispose();
 
--- a/src/FluentAvalonia/UI/Windowing/Win32/Win32WindowManager.cs
+++ b/src/FluentAvalonia/UI/Windowing/Win32/Win32WindowManager.cs
@@ -23,7 +23,10 @@ internal unsafe class Win32WindowManager
 
 #if NET5_0_OR_GREATER
         _appWindowRegistry.Add(Hwnd, this);
-        _wndProc = (nint)(delegate* unmanaged<HWND, uint, WPARAM, LPARAM, LRESULT>)&WndProcStatic;
+
+        // Apparently...nint and void* aren't blittable types to the mono-wasm compiler
+        // so the function pointer here needs to use IntPtr
+        _wndProc = (nint)(delegate* unmanaged<IntPtr, uint, IntPtr, IntPtr, IntPtr>)&WndProcStatic;
 #else
         _wndProc = Marshal.GetFunctionPointerForDelegate(WndProc);
 #endif
@@ -396,14 +399,14 @@ internal unsafe class Win32WindowManager
 
 #if NET5_0_OR_GREATER
     [UnmanagedCallersOnly]
-    private static LRESULT WndProcStatic(HWND hwnd, uint msg, WPARAM wParam, LPARAM lParam)
+    private static IntPtr WndProcStatic(IntPtr hwnd, uint msg, IntPtr wParam, IntPtr lParam)
     {
-        if (_appWindowRegistry.TryGetValue(hwnd, out var wnd))
+        if (_appWindowRegistry.TryGetValue((HWND)hwnd, out var wnd))
         {
-            return wnd.WndProc(hwnd, msg, wParam, lParam);
+            return wnd.WndProc((HWND)hwnd, msg, (WPARAM)wParam, (LPARAM)lParam);
         }
 
-        return 0;
+        return (IntPtr)0;
     }
 
     private static Dictionary<HWND, Win32WindowManager> _appWindowRegistry =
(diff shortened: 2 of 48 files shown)


robloo:
Wow, I just see this PR. Awesome work!

> Dealt with reduced Avalonia public API...which is making me absolutely lose my mind...

I was worried about that after I saw some of the breaks looking at the code. Should we discuss this more upstream about adding a few back? FluentAvalonia is a very good use case IMO and this project won't be the only one struggling.

src/FluentAvalonia/UI/Controls/MenuFlyout/MenuFlyoutInteractionHandler.cs
```diff
@@ -70,9 +70,10 @@ public virtual void Attach(FAMenuFlyoutPresenter menu)
             window.Deactivated += WindowDeactivated;
         }
 
-        if (_root is TopLevel tl && tl.PlatformImpl != null)
-            tl.PlatformImpl.LostFocus += TopLevelLostPlatformFocus;
-
+        // v2p6.1: Don't have access to this anymore...
+        //if (_root is TopLevel tl && tl.PlatformImpl != null)
+        //    tl.PlatformImpl.LostFocus += TopLevelLostPlatformFocus;
```
robloo, line 75: I wonder if we should push to add Deactivated from WPF:

https://learn.microsoft.com/en-us/dotnet/api/system.windows.window.deactivated?view=windowsdesktop-7.0

This seems like necessary information to have.

amwx: `Deactivated` is available on `WindowBase`.

Actually, thinking about this, I don't think this is actually important. This code was copied from upstream and then modified for `FAMenuFlyout`, but FAMenuFlyout is always light-dismiss so any deactivation of the top level should end up closing the flyout anyway. This file needs a refactor anyway, I just haven't felt like doing it yet.

src/FluentAvalonia/UI/Windowing/Win32/Win32WindowManager.cs
```diff
@@ -140,13 +139,18 @@ private LRESULT WndProc(HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam)
...
         }
 
-
         return CallWindowProcW(_oldWndProc, hWnd, msg, wParam, lParam);
     }
 
+    private double GetScaling()
+    {
+        // This is stupid
+        return _window.Screens.ScreenFromWindow(_window.PlatformImpl).Scaling;
```
kekekeks, line 148: RenderScaling is exposed from IRenderRoot

robloo:
@timunie I don't have any specific API needs yet on my end. I'm only trying to support where I can. Thanks for reaching out here!

amwx:
> @amwx we are really interested in hearing what makes you struggeling. Probably some API changes needs to be tweaked or added back in. Can you share a list of thing you want to have back in public API? In the best case with a short reason where it's not obvious.
> 
> I'll share it with the team to discuss.
> 
> Thank you 🙏
> 
> /cc @robloo probably you also have some API you're missing?

Of everything that was lost recently, I think `RenderScaling` is the only remaining one that really needs to be fixed. Most users of Avalonia probably won't need this, but it is something control authors might need, so I'm sure I'm not the only one. My hack of going through Screens, or the suggestion above casting to `IRenderRoot` isn't ideal. If there's some platform that wouldn't work for this, I'd suggest returning double.NaN or another sentinel value or even a `bool TryGetRenderScaling(out double scale)` on `TopLevel` - something much easier to work with and intuitive than a rarely used interface and limiting the API to the "lowest common denominator" of all supported platforms.

Thinking on my other use cases of now hidden APIs, they're probably not necessary or have better alternatives

PR #346, opened 2023-04-23
Fixes for TabView and TabViewWindowingSample

manfromarce:
- Fixes null reference exception when launching TabViewWindowingSample from the main sample app
- Selects first tab by default because it wasn't working reliably anymore
- Prevent opening a new window when there is only one tab and it's dropped outside
- Make the arrow in scroll buttons centered (it was aligned top-left)


amwx (approved):
In the interest of getting the TabView fixes into preview7, I'll accept this, though please note I've asked for a hold on PRs relating to the sample app (mentioned in v2 Roadmap pinned issue) until after the new Sample app is released

PR #352, opened 2023-04-27
[WIP] v2 Preview 7.1

amwx:
Some changes and fixes, mostly styling related:
- Switched to FABorder or FAContentPresenter in DatePicker, TimePicker, DropDownButton, and SplitButton and all "Overlay" controls (flyouts, popups, ContentDialog, and TaskDialog when in overlay mode)
  - NOTE: This only applies to the root border that has the BorderBrush/Thickness applied. To reiterate from the preview7 PR, the changes to FAContentPresenter require no changes on your part. The change to FABorder only requires and update if you have custom styles for one of these controls (just have to update from Border to ui|FABorder, where ui is your xmlns prefix)
- Removed the foreground setter on the Header ContentPresenter for `NumberBox`. 
- Fixed an issue where a NavigationViewItem without an Icon should show the content aligned as if an Icon was present. Issue was caused by `{Binding}` in a ControlTemplate is still being applied with LocalValue
- Added the new Scroll-related properties to ListBox/TreeView
- Removed the obsolete APIs from `FluentAvaloniaTheme`
  - `RequestedTheme`, `RequestedThemeChanged`, and `InvalidateThemingFromSystemThemeChanged` are now removed
  - Also removed the two constructors that required parameters as they're no longer necessary
- Cut down on a couple warnings
- Fixed issue with NumberBox caused by switch from DirectProperty -> StyledProperty

Known issues caused by things upstream:
- ComboBox won't display its last item (will be fixed by 11127)
- CommandBarButtons Label text is left aligned (if label is displayed on bottom) - related to 11125 and will be fixed by 11066

Note: branch is not yet on a CI build, I'm going to wait for at least 11127 before merging and releasing a preview7.1

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
-    <AvaloniaVersion>11.0-preview7</AvaloniaVersion>
+    <AvaloniaVersion>11.0.999-cibuild0033818-beta</AvaloniaVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="Avalonia" Version="$(AvaloniaVersion)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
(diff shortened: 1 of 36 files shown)


amwx:
@RobertBeekman Thanks for finding that - fixed it. Once again I forgot to remove the fields after switching from DirectProperty -> StyledProperty. 

I've also gone ahead and just pushed out the CI build change that I'm on. That should also fix DevTools since that has the DevTools fix (Dev tools is broken in preview7)

amwx:
Preview 7.1 was just pushed to nuget - (most likely) requires CI Build 0034035 (most recent to this post). 

Contains everything listed above, as well as #354 and #355.

PR #361, opened 2023-05-10
Preview8

### LINKED ISSUE #360: Editable `FAComboBox` TextPresenter wrong colour
**Describe the bug**
The TextPresenter of `FAComboBox` does not follow theme correctly.

**Screenshots**
Screenshot 2023-05-08 at 20 13 26

 - OS: macOS
 - FluentAvalonia Version: 2.0.0-preview7.2
 - Avalonia Version: 11.0.999-cibuild0034035-beta

amwx:
Updates to Avalonia 11.0-preview 8 along with these fixes:

- Fixed issue with `FAComboBox` textbox foreground. This was caused by `TargetNullValue` not respecting DynamicResource and `PlaceholderForeground` not being set by default (following WinUI). Fixed by including a Setter for `PlaceholderForeground` (fixes #360)
- Fixed issue with `FAComboBox` where Placeholder text (watermark) was visible when it wasn't supposed to be
- Made adjustments to `FAMenuFlyoutPresenter`, `TabView`, and `SettingsExpander` to align with new ItemsControl API for heterogeneous containers, see below

From 7.2- fixed issue with SettingsExpander when trimming is enabled

BONUS!
In preview5, the ability to use a DataTemplate to create a container for some controls was lost with the new ItemsControl. The most recent changes that allowed different container types made it once again possible to do this. For `TabView` and `SettingsExpander`, it is now (again) possible to do this (same applies to SettingsExpander):

```xaml
<ui:TabView>
    <ui:TabView.ItemTemplate>
        <DataTemplate DataType="myType">
            <ui:TabViewItem Header="{Binding MyHeader}" Content="{Binding MyContent}" ... />
        </DataTemplate>
    </ui:TabView.ItemTemplate>
</ui:TabView>
```

I hacked around that limitation in FAMenuFlyoutPresenter by force creating all containers on the flyout and then passing them to the presenter when the flyout opened (as this was a more complicated scenario), so technically this isn't as affected. I haven't undone that hack yet, will do that in a future PR.

Note that you don't have to do this, and can use the DataTemplate only for the content, the item generator will still create a container for you and use your data template for the Content only. DataTemplates can also be supplied in `DataTemplates` for implicit lookup and not just on `ItemTemplate` property of the ItemsControl.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
-    <AvaloniaVersion>11.0.999-cibuild0034035-beta</AvaloniaVersion>
+    <AvaloniaVersion>11.0-preview8</AvaloniaVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="Avalonia" Version="$(AvaloniaVersion)" />
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -4,7 +4,7 @@
     <Description>Control library focused on fluent design and bringing more WinUI controls into Avalonia </Description>
     <PackageTags>c-sharp;xaml;cross-platform;dotnet;dotnetcore;avalonia;avaloniaui;fluent;fluent-design</PackageTags>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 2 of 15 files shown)


Tryanks:
I'm glad to see that depending on the upstream fix, you finally fixed the TabView issue in preview8. Now that the #344  I had raised is really no longer useful, I should change my project back to the most classic use. Good job, thank you!

PR #375, opened 2023-05-25
FAMenuFlyout refactoring

### LINKED ISSUE #372: Crash when Settings IconSource of MenuFlyoutSubItem
**Describe the bug**

The following code crashes at runtime.

```csharp
var copyToMenuSubItem = new MenuFlyoutSubItem();
copyToMenuSubItem.Text = "Copy To";
copyToMenuSubItem.Tag  = 1;
newIcon = new FontIconSource()
{
  Glyph      = "\uF413",
  FontFamily = symbolFont,
  FontSize   = symbolFontSize,
  Foreground = iconBrush,
};
copyToMenuSubItem.IconSource = newIcon;

flyoutItems.Add(copyToMenuSubItem);
```

```
"Object reference not set to an instance of an object." System.NullReferenceException

   at FluentAvalonia.UI.Controls.MenuFlyoutSubItem.OnPropertyChanged(AvaloniaPropertyChangedEventArgs change)\r\n
   at Avalonia.Animation.Animatable.OnPropertyChangedCore(AvaloniaPropertyChangedEventArgs change)\r\n
   at Avalonia.AvaloniaObject.RaisePropertyChanged[T](AvaloniaProperty`1 property, Optional`1 oldValue, BindingValue`1 newValue, BindingPriority priority, Boolean isEffectiveValue)\r\n
   at Avalonia.PropertyStore.EffectiveValue`1.SetAndRaiseCore(ValueStore owner, StyledProperty`1 property, T value, BindingPriority priority, Boolean isOverriddenCurrentValue, Boolean isCoercedDefaultValue)\r\n
   at Avalonia.PropertyStore.ValueStore.SetLocalValue[T](StyledProperty`1 property, T value)\r\n
   at Avalonia.PropertyStore.ValueStore.SetValue[T](StyledProperty`1 property, T value, BindingPriority priority)\r\n
   at FluentAvalonia.UI.Controls.MenuFlyoutSubItem.set_IconSource(IconSource value)\r\n
   at xxx
```

**Screenshots**

 - OS: Windows 10
 - FluentAvalonia Version 2.0-preview8
 - Avalonia Version 11.0-preview8

**Additional context**

amwx:
Quite a few changes in and around FAMenuFlyout and related controls, partly because it needed it, and partly because of some coming API changes from upstream

`FAMenuFlyout`:
- Separated Items vs ItemsSource properties for FAMenuFlyout. NOTE: Unlike most ItemsControls, Items does NOT function as a view over the ItemsSource when set. ItemCollection doesn't have a public ctor so I can't use it and I don't feel like writing my own version here
- Added `FlyoutPresenterTheme` property to FAMenuFlyout
- Returned menu item generation back to FAMenuFlyoutPresenter (as was before preview 5 and the ItemsControl changes)

`FAMenuFlyoutPresenter`:
- Control no longer derives from `MenuBase` and is now just an ItemsControl
- Removed the InteractionHandler copied from upstream. All item pointer/key logic is now handled (and simplified) in FAMenuFlyoutPresenter

`MenuFlyoutItem` and related:
- Removed usage of `IMenuItem` and any related interfaces 

`MenuFlyoutSubItem`:
- Added Items vs ItemsSource properties. Same as above, Items is not a view over ItemsSource, but otherwise function as expected
- Removed `MenuFlyoutSubItemTemplate` as its no longer necessary. Just use a DataTemplate and create the MenuFlyoutSubItem that way
- Fixed #372, ensured TemplateSettings was initialized


robloo:
> ItemCollection doesn't have a public ctor so I can't use it and I don't feel like writing my own version here

I'm surprised this class isn't available for derived controls but I guess WPF is the same. Instead of writing your own version though, you could just copy/paste from upstream and rename it. It doesn't matter on my end though as ItemsSource is enough.

Alternatively, I wonder if this is a worth a discussion upstream. ItemCollection could be the base class used in other cases like this and it seems to make sense.

amwx:
> > ItemCollection doesn't have a public ctor so I can't use it and I don't feel like writing my own version here
> 
> I'm surprised this class isn't available for derived controls but I guess WPF is the same. Instead of writing your own version though, you could just copy/paste from upstream and rename it. It doesn't matter on my end though as ItemsSource is enough.
> 
> Alternatively, I wonder if this is a worth a discussion upstream. ItemCollection could be the base class used in other cases like this and it seems to make sense.

To be fair, most items related controls will derive from `ItemsControl` where `ItemCollection` is automatically created for you. I'm doing something special here since this is a flyout, and is probably not a common enough case to support opening it up to a public API.

PR #381, opened 2023-06-01
v2-Release Candidate 1

amwx:
As of this PR, the FluentAvaloniaSamples app is now considered deprecated. Now that we're in release candidate territory, the new sample app will be coming very soon (there's not much left I have to finish up, I've just been waiting for breaking changes from upstream to stop). There is a notice on the home page of the app that indicates this. Please do NOT use it to test and report bugs from. Since its replacement is coming soon, I opted not to fully update it to the most recent changes, and some stuff may not work. (Unless something major comes up that blocks it, soon means no later than early next week)

<h3>Breaking Changes</h3>

- As `AvaloniaLocator` is now hidden from public API, FATheme is no longer registered there. In order to gain a reference to it, if needed, you will have to manually query the Application styles for it.
- I've reverted the addition of FAContentPresenter I added previously in preview8. Instead I've wrapped the normal ContentPresenter in FABorder to handle the BackgroundSizing issue. (API removal upstream contributed to this and this isn't worth a discussion to get it re-added). This affects Button, ToggleButton, RepeatButton, SplitButton, DatePicker, and TimePicker. If you're using custom styles with these controls, you may have to update them for correct results.
- `BitmapIcon`/`BitmapIconSource` is currently disabled pending a way to continue Skia rendering with it
- `ColorWheel` and `ColorTriangle` of the FAColorPicker do not render, pending a way to continue using Skia. FAColorPicker needs a complete refactoring anyway

Once this releases, please do let me know if something isn't working quite right. While the latest API reductions upstream weren't contentious, they were tedious to deal with and its possible something may have gotten messed up as a result. Having everything tied to a TopLevel and trying to figure out how to work with that when you're a desktop-centric person is an interesting challenge.

Additional changes/fixes for rc1 are in the PRs since preview 8.


danielmayost:
It's worth checking if the `TaskDialog` flickering issue has been solved in https://github.com/AvaloniaUI/Avalonia/pull/11552.

PR #387, opened 2023-06-04
FAControlsGallery 

amwx:
You've probably seen me mention it a few times over the last several months and its finally here. The new sample/demo app for FluentAvalonia. I'll start with the overview demo video so you can get a look and then go into everything...

https://github.com/amwx/FluentAvalonia/assets/40413319/57894994-e74a-4981-abee-d130ef98cb69

## About the new sample app
There's lots to talk about here, as you can probably see in the video. The core of the app is still the same, but there's been a lot of refreshments to the UI to add color and animation to the UI experience. I started this back in February and have been working here and there on this and a lot of this is possible now because of the advancements and new features in Avalonia 11.

Something that's not shown in the video is that I've also included non-desktop projects too (Android, Browser, and iOS). There are some caveats here though. These are more proof of concepts rather than fully supported ways to use the demo app. The perf isn't there to support the animations so page transitions aren't as smooth. Also, I will not be able to ensure anything with the iOS project as I do not have a mac to enable testing of that. However, the android and Browser projects I can at least run, and even got it working on my phone so I know those are good to go. For full support and the best experience, the demo app is best used on a desktop environment.

Here's what it looks like in the Android emulator. The NavigationView reverts to a Left-Minimal normal mode to save screen space here. The old app was already adaptive to screen size and that was retained as much as possible.
image

## Control Definition Overlay
This was something that was in the old app, but needed some sprucing up. Previously it was kind of bland, so I've organized it a little better, with some colored icons, and added the Template Parts if a control requires specific elements in its template. This should be available for all controls, Avalonia or FA. 

## Design Guidance Section
This section (the icon above Settings in the NavView) I took from the WinUI 3 Gallery. This section replaces the old Resources page in the old app (had the DataGrid of all the resources. I wanted to improve that experience but found it hard to come up with something nice and happened to catch that Microsoft had updated their WinUI 3 gallery with these nice new pages that help with Fluent design guidance. There are 3 pages I've included:

### Typography
Includes the type ramp for using Segoe UI Variable font in windows (Fluent) design. These are available as ControlThemes for TextBlock. I've also amended this page to add info about `TextVerticalAlignmentOverrideBehavior` for cross-platform scenarios when Segoe UI isn't available.

### Icons
The Icons page contains a list of all the icons found in the FluentAvalonia.ttf font that's included with this package. A lot of these are very similar to Microsoft's Segoe Fluent Icons (including code points), but this font is built from the open source FluentUI System Icons. The list is searchable and clicking on an icon will open a TeachingTip with information, including Xaml and C# codepoints for usage. A similar list of icons is found in the IconElement page for the Symbol enum, which is subset of the full icon list. 

### Colors
And finally, the Colors page. Gives a breakdown of all the primary colors used in Fluent v2 and some guidelines on when to use them. 

## New Icons and Images
Just a little refresh of all the images and the FA icon for v2. I felt the old version of the app was very bland and definitely needed some color and I think that's been accomplished. The icons for the controls were all adapted/inspired from the images used in the WinUI 3 gallery, however, they are my own designs.

## Links
Quite a few places have links that will launch a webpage (like to this repo on github or the docs site). As Avalonia doesn't have a launcher API yet, all of these are try-catched to ensure the app doesn't crash if you click on it when on a platform that doesn't support `Process.Open`. A dialog will pop up and give you a copy-able link instead. In the future when Avalonia has a proper launcher API, this can be corrected.

Ok, now there's two things that probably caught your eye in the video that you're probably like "well are you gonna tell us about that?" 

## New Splash Screen
Users of AppWindow can take advantage of the modern splash screen API I provide there, which has 3 possible displays (simple app name, app icon, and custom content). The AppWindow page let's you see all three, but I've thought why not go big here. The new FA icon is animated here using the new Composition APIs in Avalonia 11.0. So it's not a first attempt at AnimatedIcon from WinUI or anything (just so you don't get any ideas), but is a nice demo of the new animation APIs we have.

## ConnectedAnimation
You'll also notice that navigating from the controls overview list page to a specific control page uses a ConnectedAnimation experience to move the control page icon from the list to the top of the page, and brings the page content in in the last few frames. This wasn't originally the plan, but with the extra time I had waiting for breaking changes to stop upstream before releasing this, I let the creativity off the leash a bit. It again is a good example of the Composition system. 

You're now probably wondering, "Can I use this?" The answer is sort-of. ConnectedAnimation is included in the FluentAvalonia project under the `.Controls.Experimental` namespace, as this is 100% an experiment. The API is based on UWP's ConnectedAnimation API. It's not implemented 100% and is not a supported component at this time. That means use at your own risk and issues and "how-to" questions opened will be ignored. Whether or not I make it a FA component in the future is still TBD. The Composition animation APIs need to be expanded a bit first.

# Known issues and other info
- Customizing the accent color may not updated without a render pass. I've already reported this upstream
- Remember that BitmapIcon and the ColorWheel and ColorTriangle views of FAColorPicker are unavailable in rc1
- The first time you open either the Control Definition Overlay or one of the sample code expanders, there may be a delay. This is caused by TextMate (from AvaloniaEdit) initializing which is very slow. 

There are some fixes in here too, but I can't remember what they are since some of them were pulled out and merged in a previous release and this is over several months of work.

The old sample app will be removed in a separate PR. I also noticed the other day the old Windowing project is still around which needs to be removed too, so will do that in one go.


amwx:
Also, the [docs site](https://amwx.github.io/FluentAvaloniaDocs/) has been updated to reflect some of the newer changes since it was first published and now has pages for each of the FA controls.

timunie:
Impressive Demo-App. ❤️

RobertBeekman:
These are extremely impressive and will serve as a great source of inspiration!
Any specific areas you're looking for feedback on?

amwx:
> Amazing work. it's strange but again there is a problem with FlowDirection like last time, in RTL mode the menus open on the other side (as if it were LTR). last time I fixed it upstream, also in the old sample app it's fine, I can't reproduce it upstream either.
> 
> **Edit:** Here is the culprit:
> 
> https://github.com/amwx/FluentAvalonia/blob/16077136661846302b4eec6ef611eea8f359ad71/samples/FAControlsGallery.Desktop/Program.cs#L20
> 
> I don't know why it causes this problem, any idea?

Thanks for reminding me, I noticed that too and just hadn't gotten the chance to look into it. I do need to remove the usage of Overlay popups, that was just from some build a while ago that had resource lookup issues with windowed popups. 

As for why overlay popups have issues, no idea. It's possible its a similar cause to last time where the transform is getting applied twice since the content never leaves the original TopLevel visual tree. I haven't looked at Avalonia's popup code in a while so I don't remember what differences there are in positioning between windowed vs popup. This is worth an issue upstream, I think.

samples/FAControlsGallery/Views/MainWindow.axaml.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace FAControlsGallery;
+
+public partial class MainWindow : AppWindow
+{
+    public MainWindow()
+    {
+        AvaloniaXamlLoader.Load(this);
+
+        this.AttachDevTools();
```
hez2010, line 20: This should be guarded under `#if DEBUG`

danielmayost:
I found the problem upstream, PR was sent, https://github.com/AvaloniaUI/Avalonia/pull/11680.

trrahul:
Demo app crashes if you close the Splash screen with complex content quickly before it finish loading.

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=FluentAvalonia
  StackTrace:
   at FluentAvalonia.UI.Windowing.SplashScreenContext.<RunJobs>d__11.MoveNext() in C:\Users\Rahul\Downloads\Compressed\FluentAvalonia-SampleAppv2\FluentAvalonia-SampleAppv2\src\FluentAvalonia\UI\Windowing\AppWindow\SplashScreenContext.cs:line 32
```

trrahul:
Sorry if these are known issues. Clicking No button in Task Dialog with Closing Deferral generates an exception. 

System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Avalonia.Base
  StackTrace:
   at Avalonia.Animation.AnimationInstance`1.DoPlayStates() in /_/src/Avalonia.Base/Animation/AnimationInstance`1.cs:line 140

hez2010:
The demo app crashes if you click "Show Definition" twice.

amwx:
Thanks to those who have reported issues:
- The issue with closing during a complex splashscreen has been fixed
- I am unable to repro the issue with TaskDialog. Given that the exception was thrown in the Avalonia animation code, it likely an issue upstream
- The crash with 'Show Definition' is throwing from within AvaloniaEdit and I'm not really sure why. I've wrapped the offending line on my end in a try/catch for now to at least prevent a complete failure (and everything still seems fine), but that's not a permanent fix

trrahul:
Tried the example source code for ContentDialog with Custom content, but the DataContext binding does not work. Raised an issue here https://github.com/amwx/FluentAvalonia/issues/393

amwx:
Merging as I think this is good to go now. 

If by chance you get a crash inside MediaContext dealing with the Compositor, please do NOT report that here. This is not something I can handle, but I am aware of it. It's too infrequent to get a repro on though. (Maybe already fixed upstream in nightly builds)

PR #396, opened 2023-06-21
fix(ContentDialog): Missing *Command and *CommandParameter binding

workgroupengineering:


amwx:
Note that for ContentDialog, commands are manually invoked in code thus this is missing intentionally, following WinUI. 

What is the issue here? It would be great if you can either open an issue for discussion first, or at least include a PR description that explains why a change is needed (the title is not a description).

workgroupengineering:
You are right.
The reason is the inability to use command bindidic via xaml.

```xaml
<ui:ContentDialog xmlns="https://github.com/avaloniaui"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
             xmlns:vm="clr-namespace:My.ViewModels"
             mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
             xmlns:ui="using:FluentAvalonia.UI.Controls"
             Padding="0, 10"
             x:Class="My.Views.Parts.SaveDialogView"
             x:DataType="vm:SaveDialogViewModel"
             PrimaryButtonText="Confirm"
             CloseButtonText="Cancel"
             PrimaryButtonCommand="{Binding Accept}"
             Title="Saving ..."
             Theme="{StaticResource {x:Type ui:ContentDialog}}"
             Content="{Binding .}"
             DefaultButton="Primary"
             >
  <ui:ContentDialog.ContentTemplate>
    <DataTemplate DataType="{x:Type vm:SaveDialogViewModel}">
      <StackPanel>
        <TextBlock Text="Description"/>
        <TextBox Text="{Binding Description}"/>
        <TextBlock Text="Path"/>
        <TextBox Text="{Binding Path}"
                 IsReadOnly="True">
          <TextBox.InnerRightContent>
            <Button Classes="accent"
                    Command="{Binding Browse}">
              <ui:SymbolIcon Symbol="Document"/>
            </Button>
          </TextBox.InnerRightContent>
        </TextBox>
      </StackPanel>
    </DataTemplate>
  </ui:ContentDialog.ContentTemplate>
</ui:ContentDialog>
```

robloo:
Note that WinUI/UWP controls only loosely follow MVVM Binding practices and are largely written with legacy concepts and architectures. In some cases you can't properly bind to controls like you could in WPF.

That said, Avalonia is even better than WPF in this area (binding) which means we probably shouldn't follow WinUI exactly. It was made by the WinDev team that didn't care to implement and test such things in all cases.

amwx:
@workgroupengineering Can you please be more specific about what your issue is? The button commands are invoking fine on my end as far as I can tell and "inability to use command binding via xaml" doesn't describe to me the problem in enough detail to figure out what the actual issue is here.

It's possible, since your snippet of code isn't enough to see how you're actually invoking the dialog, you may need to set the DataContext on the ContentDialog to your ViewModel, which is not done automatically (and is by design) and so it may still be null and thus the Command binding fails - this you can check with devtools when the dialog is open (it will be in the VisualLayerManager->OverlayLayer under the Window its launched from.

You could also try TaskDialog, as its a newer API not based on anything WinUI directly.

workgroupengineering:
> It's possible, since your snippet of code isn't enough to see how you're actually invoking the dialog

Code is following:

### ViewModel
```csharp
internal class SaveDialogViewModel : ViewModelBase
{
    private string? description;
    private string? path;

    public string? Description { get => description; set => SetProperty(ref description, value); }

    public string? Path { get => path; set => SetProperty(ref path, value); }

    [Avalonia.Metadata.DependsOn(nameof(Description))]
    [Avalonia.Metadata.DependsOn(nameof(Path))]
    public bool CanAccept(object? parameter) =>
        !string.IsNullOrWhiteSpace(description)
        && !string.IsNullOrWhiteSpace(path);

    public async void Accept(object? parameter)
    {
        if (File is null && Services.FsService.GetStorageProvider() is IStorageProvider storage)
        {
            var file = await storage.TryGetFileFromPathAsync(new System.Uri(path!));
            File = file;
        }
        await Task.CompletedTask;
    }

    public IStorageFile? File { get; private set; }

    public async void Browse(object? parameter)
    {
        if (Services.FsService.GetStorageProvider() is IStorageProvider storage)
        {
            var result = await storage.SaveFilePickerAsync(new()
            {
                Title = "Save Progject",
                FileTypeChoices = Services.FsService.ProjectFileTypes,
            });
            if (result is IStorageFile f)
            {

                Path = f.Path.ToString();
                File = f;
            }
        }
    }
}
```

### Invoke Dialog

```csharp
            var viewModel = new SaveDialogViewModel()
            {
                Path = this.Path,
            };
            var dialog = new SaveDialogView()  // SaveDialogView -> ContentDialog
            {
                DataContext = viewModel,
            };

            var result = await dialog.ShowAsync();
```


> You could also try TaskDialog, as its a newer API not based on anything WinUI directly.

`TaskDialog` isn't suittable for me in MVVM scenarios.

From what I've seen in `ContentDialog` the binding/assignment of the *Commands to the respective buttons is totally missing.

timunie:
```diff
var dialog = new SaveDialogView()  // SaveDialogView -> ContentDialog
{
    DataContext = viewModel,
+   PrimaryButtonCommand = myCommand,
+   PrimaryButtonCommandParamter = anyObject
};
```

this works for me. In your own impl, you can maybe setup this in the constructor to reduce the need to add it over and over again.

workgroupengineering:
> ```diff
> var dialog = new SaveDialogView()  // SaveDialogView -> ContentDialog
> {
>     DataContext = viewModel,
> +   PrimaryButtonCommand = myCommand,
> +   PrimaryButtonCommandParamter = anyObject
> };
> ```
> 
> this works for me. In your own impl, you can maybe setup this in the constructor to reduce the need to add it over and over again.

Don't work for me, Accept is always enabled.  I have attached modified version of `FASandbox` which demonstrates the issue.
`CanAccept` and `Accept` are never called because `PrimaryButtonCommand` is never assigned or binding to `PrimaryButton.Command`.


[FASandbox.zip](https://github.com/amwx/FluentAvalonia/files/11914859/FASandbox.zip)

amwx:
> Don't work for me, Accept is always enabled. 

This is the behavior of ContentDialog - you can check WinUI and you'll get the same result. If you want one of the buttons to be disabled, there's a separate `bool` property for that. This is just how ContentDialog was developed and I'm keeping FA's version inline with the behavior of WinUI.

As I said, TaskDialog is probably a better choice, as I know it does pass the Command and parameter on to the actual button instead of manually invoking like ContentDialog. I'm not sure why it's not suitable for your needs, but it certainly cannot be any worse than ContentDialog; and, it still functions similarly (not tested code, written from memory):

```xaml
<ui:TaskDialog>
    <ui:TaskDialog.ContentTemplate>
        ...
     </ui:TaskDialog.ContentTemplate>

    <ui:TaskDialog.Buttons>
        <ui:TaskDialogButton Text="Accept" Command="{Binding Accept}" ... />
        ...
     </ui:TaskDialog.Buttons>
</ui:TaskDialog>
```

Closing as ContentDialog's behavior will remain the same, and alternative solutions have been offered.

workgroupengineering:
> > Don't work for me, Accept is always enabled.
> 
> This is the behavior of ContentDialog - you can check WinUI and you'll get the same result. If you want one of the buttons to be disabled, there's a separate `bool` property for that. This is just how ContentDialog was developed and I'm keeping FA's version inline with the behavior of WinUI.
> 
> As I said, TaskDialog is probably a better choice, as I know it does pass the Command and parameter on to the actual button instead of manually invoking like ContentDialog. I'm not sure why it's not suitable for your needs, but it certainly cannot be any worse than ContentDialog; and, it still functions similarly (not tested code, written from memory):
> 
> ```
> <ui:TaskDialog>
>     <ui:TaskDialog.ContentTemplate>
>         ...
>      </ui:TaskDialog.ContentTemplate>
> 
>     <ui:TaskDialog.Buttons>
>         <ui:TaskDialogButton Text="Accept" Command="{Binding Accept}" ... />
>         ...
>      </ui:TaskDialog.Buttons>
> </ui:TaskDialog>
> ```
> 
> Closing as ContentDialog's behavior will remain the same, and alternative solutions have been offered.

I respect the decision to close the PR, but I don't agree.

The issue is also present in WinUI 3, see [here](https://github.com/microsoft/microsoft-ui-xaml/issues/8563).

If Avalonia had followed WPF steadfastly, Avalonia wouldn't be what it is today.

PR #399, opened 2023-06-24
Fix TextBoxDeleteButtonStyle

robloo:
The TextBox's InnerRightContent clear/delete button has an issue with Foreground color in dark mode. This is set using a control theme (shared several places) and currently appears as below:

**TextBox**

image

**NumberBox**

image

As can be seen above this should be white instead of black in dark mode. This PR fixes that so the foreground of the SymbolIcon is correct on initial load into dark theme and then stays correct across theme variant changes.

image


amwx (approved):
Thanks!

PR #411, opened 2023-07-03
Upgrade Avalonia to 11.0

RobertBeekman:
Not much changed this build, but my client application did fail to open so I took a quick look.
All I did was remove `AreVerticalSnapPointsRegular` and added `RoutedEventArgs e` to `OnLoaded` and `OnUnloaded` overrides.
I didn't spend too much time on the vectors since they're in experimental code.

Related PRs:
https://github.com/AvaloniaUI/Avalonia/pull/11752
https://github.com/AvaloniaUI/Avalonia/pull/11768
https://github.com/AvaloniaUI/Avalonia/pull/11828

### DIFF
--- a/src/FluentAvalonia/UI/Controls/TabView/TabView.cs
+++ b/src/FluentAvalonia/UI/Controls/TabView/TabView.cs
@@ -131,7 +131,7 @@ public partial class TabView : TemplatedControl
 
         UpdateListViewItemContainerTransitions();
 
-        OnLoaded();
+        OnLoaded(e);
     }
 
     protected override Size MeasureOverride(Size availableSize)
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <PropertyGroup>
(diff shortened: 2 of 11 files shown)


IsaacMarovitz:
Anxious to see this merged :eyes:

amwx:
Just to check in here, unfortunately 11.0 dropped at a bad time and it may not be until next week I can get through this (and the other PRs/bug fixes) to release FAv2.

However, thanks for the assist (same to any others that have open PRs)!

src/FluentAvalonia/UI/Controls/TabView/TabView.cs
```diff
@@ -131,7 +131,7 @@ protected override void OnApplyTemplate(TemplateAppliedEventArgs e)
 
         UpdateListViewItemContainerTransitions();
 
-        OnLoaded();
+        OnLoaded(e);
```
robloo, line 134: This probably isn't correct passing a new TemplateAppliedEventArgs to the OnLoaded method. It only works since they are both RoutedEventArgs but logically it doesn't make sense to do this.

This might be a wider bug here as OnLoaded() should be invoked directly by the control base class itself as part of its lifetime. I wonder if this is a hack-fix for something.

amwx (approved):

Issue #409, opened 2023-07-02
Windows `Automatic` Accent color may have undesired effect


huaxing-yuan:
This is not really a bug, I haven't met this weird behavior before setting `Accent Color` to `Automatic` by curiosity😅.

Like below screenshot (using fluentavalonia example app), after set app theme to Dark, it becomes Light several seconds later (because mt default windows app theme is Light)
![image](https://github.com/amwx/FluentAvalonia/assets/37816968/5e902fb5-03ab-41cd-8188-060254fb6314)

After looking into the code, When `Accent Color` is set to `Automatic`, `FluentAvaloniaTheme.OnPlatformColorValuesChanged()` is triggered periodically. then it overrides `RequestedThemeVariant` to Windows theme (Light). So if your app theme is set to Dark, now surprise, it becomes to Light automatically.

## Workaround:
After reading the code `FluentAvaloniaTheme.OnPlatformColorValuesChanged()` , I have replied some logic:
- Introduces 3 possible values to app theme: System, Dark and Light.
- When user chooses System Theme, `PreferSystemTheme` is set to True. With this value, application follows windows theme changes.
- When user chooses either Dark or Light Theme, `PreferSystemTheme` is set to false, and RequestedThemeVariant is assigned. With this value, application theme is fixed regardless windows settings.
- 
```csharp
  if (AppSettings.Instance.ThemeMode == AppSettings.SystemThemeString)
  {
      App.FluentAvaloniaTheme.PreferSystemTheme = true;
  }
  else
  {
      var themeMode = AppSettings.Instance.ThemeMode == FluentAvaloniaTheme.DarkModeString ? ThemeVariant.Dark : ThemeVariant.Light;
      Application.Current.RequestedThemeVariant = themeMode;
      App.FluentAvaloniaTheme.PreferSystemTheme = false;
  }
```

P.S: I started last year the migration of my application to a cross-platform solution (which heavily uses WPF and WCF). Avalonia and FluentAvalonia work great and saved me A LOT OF work and time on UI ! Thanks very much for the work!

PR #417, opened 2023-07-06
Update sample app to manage Default/Dark/Light theme taking account `PreferSystemTheme`

huaxing-yuan:
Update sample app to manage Default/Dark/Light theme while taking account `FluentAvaloniaTheme.PreferSystemTheme`.
https://github.com/amwx/FluentAvalonia/issues/409

With this update:
- If user chooses either Light and Dark theme, the theme will be fixed regardless system theme changes.
- If user chooses Default, then app theme will be synchronized with system theme


workgroupengineering:
I tested this PR with #411 and #420. If #417 is applied, theme switching does not work.

amwx:
I don't think this is the correct solution to the problem. `ThemeVariant.Default` has a very specific meaning in Avalonia and using to denote the system theme I think is a bad practice.

I need to take a closer look at your actual issue (not sure I really understand what's going on tbh) and see if I can come up with a better way. It may be this is an obscure issue in FATheme not properly handling system stuff.

huaxing-yuan:
Thanks for the reply.

To reproduce the issue, we can launch the sample application, set it to dark mode. Then go to Windows and set default app mode to light. We can see sample application changes its theme to light, regardless we have chosen dark.

In my own application, I haven't used `ThemeVariant.Default`, but an enum to store user preferred themes between "System", "Dark" and "Light" with a mechanism to handle user settings by taking account both `PreferSystemTheme` and `RequestedThemeVariant`. The idea was: if user chooses a specific theme, then we need to set `PreferSystemTheme` to `false` so that system changes won't affect application. Do you think it's a more proper solution ?

Anyway, I was just too lazy to add `ThemeVariant.Default` in the list without knowing its specific meaning.

amwx:
Oh I see, so this is more an issue with the sample app rather than an actual issue. 
In that case, I think the best course of action here is to
- change the ViewModel to use `string`s instead of `ThemeVariant`, with the options "System", "Light", and "Dark"
- Change `CurrentAppTheme` to also be a string, and in the setter:
  - if `System` is set, set `PreferSystemTheme` to true, which should automatically invalidate the current theme to match the system
  - if `Light` or `Dark` is set, ensure `PreferSystemTheme` is false and set the application's theme variant.

I did notice some issues within FluentAvaloniaTheme though, including an event handler leak with this, so if you don't mind, I'll fix up FATheme and push those changes to this branch within the next couple days.

amwx:
I've just merged in the fixes for FATheme (#434), if you'd like to make the suggested changes feel free to go ahead-you can either undo the current changes in the PR and keep it in this one, or you can close this and do a new PR, whichever you prefer.

huaxing-yuan:
Thanks, I'll update this PR and test your fix. (sorry for the late reply, I'm in vocation mode recently)

huaxing-yuan:
Updated as discussed.
Tested locally on my Windows Laptop, but unable to test on Linux and Mac from here :(

samples/FAControlsGallery/ViewModels/SettingsPageViewModel.cs
```diff
@@ -10,30 +10,58 @@ namespace FAControlsGallery.ViewModels;
 
 public class SettingsPageViewModel : MainPageViewModelBase
 {
+    readonly FluentAvaloniaTheme faTheme;
```
amwx, line 13: minor: `private readonly FluentAvaloniaTheme _faTheme;`, and move it to the bottom with the other fields

amwx (approved):
Generally looks good, thanks! If it works on Windows, I see no reason why it wouldn't work on another OS.

samples/FAControlsGallery/ViewModels/SettingsPageViewModel.cs
huaxing-yuan: ok, done.

huaxing-yuan:
Done, I've tested the first commit on Windows, Ubuntu and MacOS, it should work on at least these 3 platforms.

PR #427, opened 2023-07-18
fix(ContentDialog): Enter key handling

### LINKED ISSUE #419: ContentDialog do not block keyboard input on background controls
**Describe the bug**
ContentDialog do not block keyboard input on background controls

**Step to reproduce**
- Run `FAControlsGallery.Desktop`
- Go to `ContentDialog` page sample
- Set Default Button to `Primary`
- Click on button `Launch ContentDialog`
- Press `Enter` key on keyboard 

**Screenshots**

 - OS: **independent**
 - FluentAvalonia Version: **master 31140cccf95f9af428d90767d9f192ed6c8492f5**
 - Avalonia Version: **11.0-rc1.1**

**Additional context**

workgroupengineering:
fixes #419

- Enable nullable for `ContentDialog`
- Added KeyFilter which marks as Handled Enter KeyEventArg when focus element is not descendant of ContentDialog
- fix small nits

### DIFF
--- a/src/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
+++ b/src/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
@@ -1,14 +1,16 @@
-﻿using Avalonia;
-using Avalonia.Controls;
+﻿#nullable enable
 using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Controls.Presenters;
 using Avalonia.Controls.Primitives;
+using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Logging;
 using Avalonia.Threading;
 using Avalonia.VisualTree;
 using FluentAvalonia.Core;
+using Avalonia;
+using HarfBuzzSharp;
 
 namespace FluentAvalonia.UI.Controls;
 
@@ -62,11 +64,11 @@ public partial class ContentDialog : ContentControl, ICustomKeyboardNavigation
         return base.RegisterContentPresenter(presenter);
     }
 
-    protected override void OnKeyUp(KeyEventArgs e)
+    protected override void OnKeyDown(KeyEventArgs e)
     {
         if (e.Handled)
         {
-            base.OnKeyUp(e);
+            base.OnKeyDown(e);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


amwx:
A couple of things:
- Do not enable nullable reference types. This is in my contributing guidelines.
- I don't like the catch all key handler. That can be dangerous if someone has a custom control or view and wants to handle the enter key in a special way and prevent the dialog from handling it. It also feels like a hack for a problem that really shouldn't exist and is rooted elsewhere
- I'm thinking using Loaded event is probably the best place to handle final initialization of the dialog (like setting focus), however, we have to be really careful about that, as the dialog can be declared in Xaml and thus Loaded can fire even when the dialog isn't open

As I said in #419, I need to understand why this is occurring (mostly why it stopped working) before figuring out what the best solution is here.

workgroupengineering:
> A couple of things:
> 
>     * Do not enable nullable reference types. This is in my contributing guidelines.
> 

It's my mistake, I revert it.


> * I don't like the catch all key handler. That can be dangerous if someone has a custom control or view and wants to handle the enter key in a special way and prevent the dialog from handling it. It also feels like a hack for a problem that really shouldn't exist and is rooted elsewhere

In my opinion, `ContentDialog` being a modal window should block all input directed to non-child controls of `ContentDialog` and not just the `Enter` key like I did.

> As I said in #419, I need to understand why this is occurring (mostly why it stopped working) before figuring out what the best solution is here.

The reasons for the issue are multiple:

- focus control cannot be set before it is loaded. 

- When `DefaultButton` is different from `None` if the `ShowAsync` method is called during a `Enter` KeyDownEvent (eg: activation via enter of a button ), when the `Enter` key is released the ContentDialog's OnKeyUp is executed which causes the closing of the `ContentDialog `.

- When `DefaultButton` is `None` and `ContentDialog` does not have a focusable control that can intercept keyboard input, keyboard input is passed to the last actived control before calling `ContentDialog.ShowDialog`.

amwx:
Ok, so I've made progress on the underlying issue. The focus code is failing because the dialog is announcing its ready to go before all the initialization completes, which means the buttons are still hidden when the focus code runs - which means it ends up failing.

Sorry to do this, but I'm going to close this and take this issue over myself so I can make sure we don't run into lifetime issues here. I'm doing this alongside adding unit tests (which is how I discovered the underlying issue), so we can be sure this issue is fully fixed and doesn't come up again.


> In my opinion, `ContentDialog` being a modal window should block all input directed to non-child controls of `ContentDialog` and not just the `Enter` key like I did.

I agree with you here, and I did try to implement that at one point, but the dialogs aren't actually windows, it's an overlay in the `OverlayLayer`. And the problem was blocking all keyboard input prevented textboxes from working, since marking key events as handled prevents the text events from running through. There's an issue/PR fix for that somewhere, but I can't find it at the moment. Ideally, Avalonia needs to do what UWP/WinUI does and completely separate the main visual tree from the overlay layer, such that routed events that originate in the overlay layer don't end up making their way to the parent window. Though this is quite complicated as you need to abstract a "Window (TopLevel)" concept (that isn't the traditional WPF style Window) that all the events would then run through so that text processing still works, UWP calls this the XamlRoot:
XamlRoot
-> Main Visual Tree (Page content)
-> PopupRoot layer (used for popups and ContentDialog)

And unfortunately, such a change like this, if the Avalonia team is even interested, would have to wait until 12.0 since it would be a massive change to how things work.

I think you tried to sort of handle this by implementing a filter to figure out if the source originated from within the dialog, but I worry constantly needing to check that will end up being a perf hit if we're constantly scanning the visual tree and I'd rather not do that.



>     * focus control cannot be set before it is loaded.
> 
>     * When `DefaultButton` is different from `None` if the `ShowAsync` method is called during a `Enter` KeyDownEvent (eg: activation via enter of a button ), when the `Enter` key is released the ContentDialog's OnKeyUp is executed which causes the closing of the `ContentDialog `.
> 
>     * When `DefaultButton` is `None` and `ContentDialog` does not have a focusable control that can intercept keyboard input, keyboard input is passed to the last actived control before calling `ContentDialog.ShowDialog`.

As I said above, I know why the first one is occurring, that's the root issue here. The other two, those are good points and I'll see what I can do about that.

Thanks for your interest in this though!

PR #429, opened 2023-07-18
Ensure UpdateTabWidths works with add INCC changes

### LINKED ISSUE #424: TabView - Incorrect tab sizes
Using v2.0.0 of FluentAvalonia, the sizes of Tabs in a TabView are occasionally not correct. This can be seen in the ControlsGallery sample. 

In any of the TabView examples that support closing tabs (though the 2nd example seems to work best), close the last tab then add a new tab *without moving the cursor after pressing add*, the Tab will be too small until the cursor moves out of the tab, at which point it will resize to the correct size. This is likely a race condition as it doesn't always exhibit this behaviour.

A bit of debugging suggests the issue is that the `UpdateTabWidths` function is being called too soon. It's being called when the TabItems collection changes, however at that point the underlying presenter panel has not been modified, hence `ContainerFromIndex` for the tab returns null and nothing is resized.

amwx:
As in #424. adding a TabViewItem would sometimes cause the Tab to be incorrectly sized until the pointer leaves the tab strip. The issue was caused by `UpdateTabWidths()` being called from the INCC handler which is before the items have been realized in the items panel, which meant the new item wasn't accounted for. PR posts the update to the Dispatcher to wait until the container is realized so it works

Fixes #424

### DIFF
--- a/src/FluentAvalonia/UI/Controls/TabView/TabView.cs
+++ b/src/FluentAvalonia/UI/Controls/TabView/TabView.cs
@@ -9,6 +9,7 @@ using Avalonia.Controls.Templates;
 using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Logging;
+using Avalonia.Threading;
 using Avalonia.VisualTree;
 using FluentAvalonia.Core;
 using FluentAvalonia.UI.Controls.Primitives;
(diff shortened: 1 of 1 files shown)


brads55:
This does appear to fix the issues I'm seeing, thanks.

PR #433, opened 2023-07-20
ContentDialog improvements

### LINKED ISSUE #419: ContentDialog do not block keyboard input on background controls

amwx:
Several much needed improvements for ContentDialog:

- Initialization logic has been refactored to ensure ContentDialog initalizes in the correct order. Previously, `SetupDialog`, which is responsible for setting up all the buttons based on the properties was called AFTER the dialog said is was ready and raised the opened event. This was leading to the issue where the buttons were still invisible and initial focus was not being applied. Now, everything is run in the correct order and when the Opened event fires, you can be sure the entire dialog is ready to go. 
  - As a result, initial focus is now correctly applied (this is done automatically). You can still override that and focus whatever you want by handling the Opened event.
  - It is not required to post to the dispatcher to get focus to work correctly
  - Also, just as a reminder, if no focusable content is found in the ContentDialog (can happen with all disabled buttons or no buttons, combined with no focusable content), focus will NOT move into the dialog and will remain on whatever the previous focus is. This is by design and is how the FocusManager works (we can't focus any control that isn't focusable)
- Fixed an issue where if the dialog was launched via a Button invoked with the enter key with a DefaultButton set, the dialog would immediately close. This is hack fixed as I think this is an underlying issue in Avalonia where if focus changes in between key events, the KeyUp is routed to the new focused control rather than the old.
- Some adjustments to the ShowAsync method
  - Added `ShowAsync(TopLevel)` overload. This should be used in cases where no ApplicationLifetime is present (like in my unit tests), as ContentDialog won't know where to launch. 
  - As a result, `ShowAsync(Window)` remains for now but will be removed in 2.1 as this is breaking change
- Some minor cleanup of some code
- Added unit tests for ContentDialog, which includes tests for the above bugs to ensure they don't regress in the future.

fixes #419

### DIFF
--- a/src/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
+++ b/src/FluentAvalonia/UI/Controls/ContentDialog/ContentDialog.cs
@@ -39,9 +39,6 @@ public partial class ContentDialog : ContentControl, ICustomKeyboardNavigation
         _secondaryButton.Click += OnButtonClick;
         _closeButton = e.NameScope.Get<Button>(s_tpCloseButton);
         _closeButton.Click += OnButtonClick;
-
-        // v2- Removed this as I don't think its necessary anymore (called from ShowAsync)
-        //SetupDialog();
     }
 
     protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs change)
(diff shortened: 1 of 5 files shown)


timunie:
> This is hack fixed as I think this is an underlying issue in Avalonia where if focus changes in between key events, the KeyUp is routed to the new focused control rather than the old.

If we have no issue to track this upstream, consider to open an issue upstream for this.

workgroupengineering:
ContentDialog Issue

When `DefaultButton` is `None`, the Enter key should do nothing

amwx:
@timunie There's a related issue already filed several months ago [here](https://github.com/AvaloniaUI/Avalonia/issues/9626)

@workgroupengineering Not entirely true. The reason the enter key is closing the dialog is because focus is set to one of the buttons (as its the only focusable content in the dialog), so the enter key is passed to the button which invokes it - the dialog itself isn't doing anything. This is correct behavior - the focused control takes priority, then the dialog steps in after that, if needed. When default button is None, the enter key should be ignored for unhandled Key events that come from within the dialog's content (like if you had a TextBox there)

Issue #228, opened 2022-11-01
Use XmlnsDefinitionAttribute to map XML namespace to CRL namespace


thexamlguy:
In doing so, this will remove all of the ugly namespace prefixes found all over the XAML, i.e.

From this:
```
<UserControl
    x:Class="Builder.ProjectView"
    xmlns="https://github.com/avaloniaui"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:fluent="using:FluentAvalonia.UI.Controls">
    <StackPanel>
        <fluent:Frame />
        <fluent:FontIcon />
        <fluent:NavigationView>
            <fluent:NavigationView.MenuItems>
                <fluent:NavigationViewItem />
			<fluent:NavigationViewItem />
	        </fluent:NavigationView.MenuItems>
        </fluent:NavigationView>
    </StackPanel>
</UserControl>
```

to this
```
<UserControl
    x:Class="Builder.ProjectView"
    xmlns="https://github.com/avaloniaui"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
    <StackPanel>
        <Frame />
        <FontIcon />
        <NavigationView>
            <NavigationView.MenuItems>
                <NavigationViewItem />
			<NavigationViewItem />
		</NavigationView.MenuItems>
        </NavigationView>
    </StackPanel>
</UserControl>
```

This can be done by adding a few lines to the Assembly.cs file in the FluentAvalonia.UI.Controls project.
```
using Avalonia.Metadata;
[assembly: XmlnsDefinition("https://github.com/avaloniaui", "FluentAvalonia.UI.Controls")]

```

amwx:
I *really* don't like when 3rd parties do this. 
1- It causes issues with the designer and I don't wanna have to triage issues relating to "why isn't this working"
2- It makes it seem the 3rd party controls are part of the main library - which they're not. This violates Xml specs, as nodes in different namespaces need their xmlns prefix
3- It could cause type name clashes. Until recently, I had a couple controls that had the same name as upstream. I've now prefixed them with 'FA' to distinguish, but this could cause other issues

I get this makes the Xaml a little cleaner and removes the need to add the xmlns stuff (which can be annoying for every xaml file), but IMO this is a bad practice.

PR #446, opened 2023-08-04
Add XML Namespace Definitions

HuJinguang:
When we add a reference to this library in an Avalonia UI XAML file, we can use the URL identification format. For example: 
`
xmlns:fluent="https://github.com/fluent-avalonia"
`

### DIFF
--- a/src/FluentAvalonia/Properties/AssemblyInfo.cs
+++ b/src/FluentAvalonia/Properties/AssemblyInfo.cs
@@ -1,3 +1,11 @@
 ﻿using System.Runtime.CompilerServices;
 
+using Avalonia.Metadata;
+
 [assembly: InternalsVisibleTo("FluentAvaloniaTests")]
+
+[assembly: XmlnsDefinition("https://github.com/fluent-avalonia", "FluentAvalonia.UI")]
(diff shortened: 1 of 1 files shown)


WidgetifyApp:
Duplicate of https://github.com/amwx/FluentAvalonia/issues/228 which has been answered as won't do I'm afraid.

HuJinguang:
amwx is right, but I'm using a different URL than the main library, which doesn't confuse people

amwx:
I still don't like the URL approach. The explicit namespaces keeps things separated so you don't get unnecessary completion suggestions because of something in another namespace. 99% of the stuff you need in this library is in the `FluentAvalonia.UI.Controls` namespace. You need one in App.axaml for .Styling, and maybe .Windowing if you need the attached property for hit test or doing a custom window style (which is ill-advised anyway). There isn't really anything in .Primitives that's useful on its own. Is it really that inconvenient to do xmlns:ui="using:FluentAvalonia.UI.Controls" over the url and then the others in the rare case you need them?

HuJinguang:
You're right, but there are still a lot of people who like to use the URL approach, so I think it's possible to set URLs for different namespaces to suit the preferences of different developers without breaking the design

Issue #458, opened 2023-08-12
AppWindow system chrome icon glyph changes


cricketthomas:
**Is your feature request related to a problem? Please describe.**

I noticed that the glyphs in AppWindow are thicker than other windows app, even though the current glyphs used are the ones you'd expect MSFT to be using on their own fluent apps.
FA app window:
![image](https://github.com/amwx/FluentAvalonia/assets/15821271/ca1a28ed-9adf-4a09-b1ea-471d4c3ea80e)
Edge:
![image](https://github.com/amwx/FluentAvalonia/assets/15821271/0ebc0a85-b85a-431a-ab2f-594d65058c7e)

**Describe the solution you'd like**

Update the glyphs in AppWindow to match other windows app. 

**Describe alternatives you've considered**

Override the icons in my own application, which may be more appropriate since this is a small style opinion.

PR #459, opened 2023-08-12
Windows chrome glyphs icons changes

cricketthomas:
This would close https://github.com/amwx/FluentAvalonia/issues/458 if this solution is appropriate.

I noticed that the system chrome icons in AppWindow were slightly larger than the ones seen in other windows apps, and it seems like windows is not using the glyphs they explicitly call "ChromeMaximize" etc.. on their own apps. I am wondering if they should be more of a copy of other fluent apps?

I dug through the [glyphs](https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-fluent-icons-font) and I wondering if these match slightly more in your opinion, if not this can be closed.

Current chrome icons:
image
Fullscreen:
image

PR version of chrome icons changes:
image
Fullscreen:
image

Comparison with a few other apps on win11 with the change:
Screenshot 2023-08-12 105848

This uses the already defined FontSize of 14 and allows updating it to 16px for the "X" symbol.

### DIFF
--- a/src/FluentAvalonia/Styling/ControlThemes/FAControls/AppWindowStyles.axaml
+++ b/src/FluentAvalonia/Styling/ControlThemes/FAControls/AppWindowStyles.axaml
@@ -22,8 +22,7 @@
         <Setter Property="Template">
             <ControlTemplate>
                 <Border Background="{TemplateBinding Background}">
-                    <Viewbox Width="14"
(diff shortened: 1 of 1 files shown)


amwx (approved):
Thanks! Kinda mad this problem had such a simple solution 😂

PR #471, opened 2023-09-03
New Control: RangeSlider

### LINKED ISSUE #274: Add RangeSelector control from Windows Community Toolkit
Control: https://github.com/CommunityToolkit/WindowsCommunityToolkit/tree/63cbb4a51bdef59083ccfc86bbcba41f966d0027/Microsoft.Toolkit.Uwp.UI.Controls.Input/RangeSelector

Docs:
https://learn.microsoft.com/en-us/windows/communitytoolkit/controls/rangeselector

amwx:
First new control in the v2 era, the RangeSlider - adapted from the WinUI community toolkit's RangeSelector.
https://github.com/CommunityToolkit/WindowsCommunityToolkit/tree/63cbb4a51bdef59083ccfc86bbcba41f966d0027/Microsoft.Toolkit.Uwp.UI.Controls.Input/RangeSelector

https://learn.microsoft.com/en-us/windows/communitytoolkit/controls/rangeselector

There were some changes to the control I've made for various reasons, but it's mostly the same control as in the WinUI toolkit.

RangeSlider
image

RangeSlider with ToolTip while dragging
image

Still TODO:
- [x] Implement ToolTip string format
- [x] Deal with some remaining ToolTip nonsense
- [x] Thumb key events
- [x] Unit tests
- [ ] Sample page in FAControlsGallery

---

API & changes
- Control is called `RangeSlider` instead of `RangeSelector`, which makes more sense for what it is
- WinUI tooltip is a component in the template (not an actual popup/tooltip) which has clipping and size issues. I've changed this version to use an actual ToolTip. There is still the ToolTip component in the template (which you can customize by re-templating the control, if you choose). Its removed from the template in OnApplyTemplate so trying to target it via styles won't work
  - I've added a property (`ToolTipStringFormat`), however, to control the string format for the text in the ToolTip, so that level of customization is pretty easy.
- There is no vertical mode (also doesn't exist in WinUI). I have to plans to implement or support a vertical RangeSlider.

Closes #274

Also fixes an issue with the Slider thumb style (this also impacts the regular Slider control), where the elevation brush was still inverted (from before we had brush transforms). You probably won't notice any difference unless you look real close, but the brush/thumb now renders correctly right side up


amwx:
Anyone who wants to test, feel free. Please, if you find an issue, check with the WinUI community toolkit to see if it repros there (and if there's a open issue about it). If its a core issue with the control, I may or may not be able to fix it.

timunie:
Well okay I have some feedback I want to share with you, but this will differ from the control you ported to Avalonia. I'll let you know anyway, leaving up to you to fully ignore it.

At the moment we only have the possibility to set either Min or Max. There is no middle thumb to move the entire Range. This has two draw backs imo:
1. If the values are close to each other, the thumbs will overlap so you cannot move the upper value easily:
image 
image

2. There is no chance to move upper and lower limit in sync. 
image

In WPF I used this control which does support 1 and 2: https://mahapps.com/docs/controls/rangeslider

There was an issue upstream (but just closed for no reason):
- https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/2884

timunie:
> This is going to be an issue with this type of control regardless. Whichever thumb was the last to be dragged (or RangeEnd if it hasn't been interacted with) is the one on top, which you'll have to move out of the way if you want the other thumb. 

The middle track could get a given MinWidth. For sure that must be respected int he calculation, but would allow to have the thumbs not beeing overlappend, if the developer / designer wants this. If that `MiddleThumbWidth` defaults to `0`, the current behavior is still the preserved. Only drawback would be, that I think the tick marks cannot be rendered, if `MiddleThumbWidth != 0`

amwx:
New Changes:

- `MinimumRange` property has been added. This allows you to specify the minimum distance between thumbs, so that you can prevent them from overlapping. This is in data units, not pixels, as expected. Note: you can still programmatically set the range closer together, this only prevents dragging too close. 
- I've updated the logic so that when you drag a thumb into the other (including if MinimumRange is set), it will then move the entire range. For ex, if you drag the RangeStart thumb into RangeEnd, it will then move both RangeStart and RangeEnd together. This applies with or without a MinimumRange.
- I've added a property to control whether the Value ToolTip displays when dragging, so you can turn it off altogether if you want.
- Improvement over the WinUI version, the ValueChanged event is now fired when RangeStart or RangeEnd is programmatically changed. And a note on this, as in WinUI version, ValueChanged is only fired after the drag is completed, intermediate changes while dragging will not raise the event - likely because of the way the control was written, and I'm not trying to reinvent the wheel here.
Also just to manage expectations, I have no plans to add or support drawing tick marks for this control.

timunie:
> MinimumRange property has been added. This allows you to specify the minimum distance between thumbs, so that you can prevent them from overlapping. This is in data units, not pixels, as expected. Note: you can still programmatically set the range closer together, this only prevents dragging too close.

This is a good idea 👍 and will make sure that one can configure it quite easily. However, I think that we should add `MiddelTrack.MinWidth` into the calculations. That way we don't need to add additional properties but the desginer can still set a MinWidth of the middle track, to have it shown even if both values are the same. But this can also be done later in time, if really needed. Let's start with what we have rn. 

Thanks for listening to feedback ❤️

PR #505, opened 2023-12-27
Frame Enhancements

MakesYT:
1. Add the NavigateFromInstance method for Frame
Causes the frame to load content represented by the instance of a Control with the specified navigation options
2. Add IServiceProvider support for Frame
Allow it to preferentially obtain instances through IServiceProvider rather than create them directly

### DIFF
--- a/src/FluentAvalonia/UI/Controls/Frame/Frame.cs
+++ b/src/FluentAvalonia/UI/Controls/Frame/Frame.cs
@@ -99,6 +99,7 @@ public partial class Frame : ContentControl
         }
     }
 
+    public IServiceProvider Services { get; set; }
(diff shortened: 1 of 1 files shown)


amwx:
1. Using a specific page/control instance for navigation seems very specific and probably means you're using Frame wrong. You'd probably be better off with a ContentPresenter and managing your navigation stack yourself if you want this.

2. Use the `NavigateFromObject` method and supply an `INavigationPageFactory` to create pages yourself using an arbitrary parameter. The methods you'll implement for INavigationPageFactory can then invoke whatever service provider you want, no need to add this to the control itself.

These changes are not likely to be merged, but thank you for taking an interest.

PR #557, opened 2024-04-27
Control Fixes

robloo:
1. Use ContentPresenter instead of ContentControl in ComboBox
     * This is what the style selectors are using and is also the correct control to use here.
 1. Update CalendarDatePicker theme
     * The calendar glyph is currently always black in dark theme -- not light. I fixed this by setting the Foreground from a style selector rather than hardcoded in the control theme. That said, I'm not sure why it wasn't being inherited correctly as it looks like it should have worked. This might be an upstream bug.
     * The CalendarDatePicker had no styles for pointerover, disabled, etc. relying entirely on the TextBox itself to provide this. This completely ignores the CalendarDatePicker resources apps might be using. Therefore, I added support for this based on upstream WinUI.
     * https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/1.5-stable/controls/dev/CommonStyles/CalendarDatePicker_themeresources.xaml

### DIFF
--- a/src/FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarDatePickerStyles.axaml
+++ b/src/FluentAvalonia/Styling/ControlThemes/BasicControls/CalendarDatePickerStyles.axaml
@@ -92,10 +92,8 @@
             </ControlTemplate>
         </Setter>
 
-        <Style Selector="^">
-            <Style Selector="^ /template/ Button#PART_Button">
-                <Setter Property="Foreground" Value="{DynamicResource CalendarDatePickerCalendarGlyphForeground}" />
-            </Style>
+        <Style Selector="^ /template/ Button#PART_Button">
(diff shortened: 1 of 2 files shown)


amwx (approved):

robloo:
> The calendar glyph is currently always black in dark theme -- not light. I fixed this by setting the Foreground from a style selector rather than hardcoded in the control theme. That said, I'm not sure why it wasn't being inherited correctly as it looks like it should have worked. This might be an upstream bug.

Note that this may be an issues fixed by https://github.com/AvaloniaUI/Avalonia/pull/15670. The work-around here shouldn't really be an issue but it might be good to revert at some point in the future (note to myself).

robloo:
Also see upstream https://github.com/AvaloniaUI/Avalonia/issues/15729 discussion

PR #558, opened 2024-04-28
Add ProgressRing

amwx:
Another little thing I've been experimenting with. This PR adds the modern style ProgressRing control to FA! 

Previous attempts have run into issues with performance around this. This version shifts the animation over the Composition thread which should significantly help with that, and keep the progress ring running smoothly even if the UI thread slows for whatever reason. Thus, the actual attempt is done directly with Skia and will require the Skia rendering backend. 
WinUI's AnimatedVisualPlayer (Lottie) also runs on the composition thread, so behavior should basically be identical. I wasn't able to make sense of the generated Lottie AnimatedVisualSource for WinUI's progress ring, so this attempt is entirely reverse engineered by watching the animation - it seems close enough to me.

https://github.com/amwx/FluentAvalonia/assets/40413319/6e8806e6-6ca8-4cc5-9f1f-b17f2a6aa3b3


robloo:
It was a good idea to use composition here to offload the animation-related changes to another thread! You also have a good eye to figure out that animation by looking just at the graphics! Going to test this out more when it's released for sure.

The only thing I notice is the animated delay going to 100 % determinate progress. I believe in WinUI determinate changes are updated instantly so a delay may be unexpected here -- especially for fast percentage changes.

amwx:
> The only thing I notice is the animated delay going to 100 % determinate progress. I believe in WinUI determinate changes are updated instantly so a delay may be unexpected here -- especially for fast percentage changes.

No, that is correct (you can see it in the WinUI 3 gallery). I almost left it out because it had to be coded in and I don't really like it, but if I did inevitably someone would want it. It is instantaneous, however, if you decrease the value (100 to 25) - the animation only plays if the value increases.

robloo:
Ah, OK. When WinUI started switching to the Lottie animations that's about when I stopped following the project closely. So I'm thinking of old functionality then.

PR #574, opened 2024-05-27
Improve RangeSlider

### LINKED ISSUE #568: StackOverflow exception in RangeSlider
*Transferred issue:*

this triggers the issue: 
1. check out this branch of FluentAvalonia: https://github.com/timunie/FluentAvalonia/blob/issueReport/
2. Run `samples/FAControlsGallery/Views/TestView.axaml`
3. You will see 2 input fields and one slider. In the input field add a number that is out of Range, for example `102`. 
4. The App crashes with the StackOverflow expection

_Originally posted by @timunie in https://github.com/AvaloniaUI/Avalonia/issues/15539#issuecomment-2107780917_

timunie:
## Descrition of changes

- Use `CoerceValue` where possible (same as Slider in Avalonia uses)
- Add property to control snap behavior `IsSnapToStepFrequencyEnabled` (not everyone may like it, same as Slider in Avalonia)
- SnapToFrequency should be handled in input events, not in `PropertyChanged` (don't touch Values from the ViewModel sent)

## To test the RangeSlider

I added a minimum sample to SandBox. Can revert this change before merging if you want me to

## Tests
Well ... Need to think about useful tests later on, if that PR gets accepted. 

## Fixed issues
Fixes #568

### DIFF
--- a/samples/FASandbox/MainWindow.axaml
+++ b/samples/FASandbox/MainWindow.axaml
@@ -1,16 +1,10 @@
-<Window xmlns="https://github.com/avaloniaui"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        xmlns:ui="using:FluentAvalonia.UI.Controls"
-        xmlns:uip="using:FluentAvalonia.UI.Controls.Primitives"
-        xmlns:core="using:FluentAvalonia.Core"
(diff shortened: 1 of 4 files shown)


timunie:
Hi @amwx 

since 2.1.0 seems to be released already, should I switch target branch?

dowdybrown:
We use the RangeSlider in our project and can't update to Avalonia 11.1.x until this pull request is approved and released. I would be happy to provide any assistance needed toward that end.

timunie:
@dowdybrown found out that the underlaying issue is solved in latest Avalonia nightly. However FA needs some updates to handle this version so we will need to wait for a fix or build FA on our own.

PR #626, opened 2024-12-01
Fix VerticalTextAlignment Override not being set when initializing

### LINKED ISSUE #616: Text alignment is not vertically centered in MacOS, Linux
**Describe the bug**
The text alignment is not vertically centered in some controls like ComboBox.
The margin between Header and Description in Expander is also a bit short.

**Screenshots**
image
_MacOS_

image
_Linux_

image
_Windows 11 for comparison_

 - OS: MacOS Sequoia, Fedora
 - FluentAvalonia Version: 2.1.0
 - Avalonia Version: 11.2.0-rc1

**Additional context**
Looks like it should be fixed by https://github.com/amwx/FluentAvalonia/issues/128, `TextVerticalAlignmentOverride` is not set(default.)

amwx:
As noted in #616, text alignment was not properly centered on non-Windows systems - even if using the `TextVerticalAlignmentOverrideBehavior` property on `FluentAvaloniaTheme`. This issue was caused because FATheme gets initialized before any properties set in Xaml are set (init in ctor). This issue dates back to when FATheme was forcibly moved to a Xaml file to accommodate AOT, so its apparently been around a while. 

I've adjusted the property setter to call the method that activates the special styles so that it calls. Note, at runtime you can only turn on the override. Setting the property to turn it off will do nothing - so it still should be treated as a start-up only property. I also adjusted one of styles to correct for `ComboBox`.

Fixes #616

Note: I've tested with a non Segoe font on Windows and it is resolved, but if someone is able to test this branch on Linux or Mac without a package, please let me know if this fixes the issue.

### DIFF
--- a/src/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml.cs
+++ b/src/FluentAvalonia/Styling/Core/FluentAvaloniaTheme.axaml.cs
@@ -1,14 +1,17 @@
 ﻿using Avalonia;
 using Avalonia.Collections;
 using Avalonia.Controls;
+using Avalonia.Controls.Presenters;
 using Avalonia.Layout;
 using Avalonia.Markup.Xaml;
 using Avalonia.Media;
 using Avalonia.Platform;
 using Avalonia.Styling;
+using Avalonia.Threading;
 using FluentAvalonia.Interop;
 using FluentAvalonia.UI.Media;
 using System.Collections.Specialized;
+using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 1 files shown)


nilaoda:
image

It still doesn’t look correct on macOS.

Jannik-M91:
This partially fixes https://github.com/amwx/FluentAvalonia/discussions/751 but I think the same fix must be done for `UseSystemFontOnWindows `which should probably call `ResolveThemeAndInitializeSystemResources()` or `ResolveWindowsSystemSettings() `when being set.

Jannik-M91:
I have tested this fix on both Windows and macOS using Avalonias Inter fonts.

On Windows it works fine, on macOS an exception is thrown because `TextVerticalAlignmentOverrideBehavior` is now applied twice, once during Init() and then again when the attribute is set.

I have implemented a possible fix here which consists of setting the default for `TextVerticalAlignmentOverrideBehavior` to Disabled and not calling `SetTextAlignmentOverrides()` during `Init()`: https://github.com/amwx/FluentAvalonia/commit/d7c2a657180c58512d6cf5fec34b093292a51b06

Apart from this the fix works fine for me on both Windows using non-Segoe fonts and on macOS so I would love to see it added with the next release.

Also having a similar fix for `UseSystemFontOnWindows` which is also not loaded during `Init()` at the moment and has no runtime setter would be appreciated.

PR #629, opened 2024-12-15
Fix: Add fallback for missing DPI-related Win32 APIs

### LINKED ISSUE #628: Application Fails to Launch on Older Windows 10 Versions Due to Unsupported API
## Describe the bug
A clear and concise description of what the bug is and any steps to reproduce it. Sample code or a minimal repro is helpful.

## Desktop/Platform
 - **OS**: Windows 10 10240, LTSC 2015
 - **FluentAvalonia Version**: 2.2.0
 - **Avalonia Version**: 11.2.1

## Error Details
```
Application: Demo.exe  
CoreCLR Version: 9.0.0  
Description: The process was terminated due to an unhandled exception.  

Exception Info: System.EntryPointNotFoundException: Unable to find an entry point named 'AdjustWindowRectExForDpi' in native library 'user32.dll'.  
   at Internal.Runtime.CompilerHelpers.InteropHelpers.FixupMethodCell(IntPtr, InteropHelpers.MethodFixupCell*) + 0x15c  
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ResolvePInvokeSlow(InteropHelpers.MethodFixupCell*) + 0x3d  
   at FluentAvalonia.Interop.Win32Interop.AdjustWindowRectExForDpi(RECT*, Int32, BOOL, Int32, Int32) + 0x2f  
   at FluentAvalonia.UI.Windowing.Win32WindowManager.HandleNCCALCSIZE(WPARAM, LPARAM) + 0x159  
   at FluentAvalonia.UI.Windowing.Win32WindowManager.WndProc(HWND, UInt32, WPARAM, LPARAM) + 0x21e  
   at FluentAvalonia.UI.Windowing.Win32WindowManager.WndProcStatic(IntPtr, UInt32, IntPtr, IntPtr) + 0x72  
```

Possible Cause:
* The application calls the `AdjustWindowRectExForDpi` API, which is only available on newer Windows 10 version.

nilaoda:
image

fix #628

### DIFF
--- a/src/FluentAvalonia/Interop/Win32Interop.cs
+++ b/src/FluentAvalonia/Interop/Win32Interop.cs
@@ -20,17 +20,20 @@ internal static unsafe partial class Win32Interop
     public static extern BOOL AdjustWindowRectEx(RECT* lpRect, uint dwStyle, BOOL bMenu, uint dwExStyle);
 
     [DllImport(s_user32, SetLastError = true)]
-    public static extern int GetSystemMetrics(int smIndex);
+    private static extern int GetSystemMetrics(int smIndex);
 
     [DllImport(s_user32, SetLastError = true)]
-    public static extern int GetSystemMetricsForDpi(int nIndex, uint dpi);
+    private static extern int GetSystemMetricsForDpi(int nIndex, uint dpi);
 
     [DllImport(s_user32, SetLastError = true)]
     public static extern BOOL SetWindowPos(HWND hWnd, HWND hWndInsertAfter, int x, int y, int cx, int cy, uint uFlags);
 
     [DllImport(s_user32, SetLastError = true)]
-    public static extern BOOL AdjustWindowRectExForDpi(RECT* lpRect, int dwStyle, BOOL bMenu, int dwExStyle, int dpi);
-
+    private static extern BOOL AdjustWindowRectExForDpi(RECT* lpRect, int dwStyle, BOOL bMenu, int dwExStyle, int dpi);
+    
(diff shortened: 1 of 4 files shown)


nilaoda:
@amwx When you have a moment, I’d appreciate your review and any feedback you might have. Thanks a lot!

src/FluentAvalonia/Interop/Win32Interop.cs
```diff
@@ -63,6 +66,33 @@ public static extern BOOL TrackPopupMenu(HMENU hMenu, uint uFlags,
     // Below is adapted from TerraFX.Interop.Windows, MIT License
     public static delegate*<HWND, int, nint> GetWindowLongPtr => &GetWindowLongPtrW;
 
+    public static int GetSystemMetricsWithFallback(int nIndex, uint dpi)
+    {
+        try
+        {
+            return GetSystemMetricsForDpi(nIndex, dpi);
+        }
+        catch (EntryPointNotFoundException)
```
amwx, line 75: It would be better to use a Windows version check rather than Try/Catch. Take a look at the `ApplyTheme` method further below, as I already have a helper class for this.

amwx (changes requested):

src/FluentAvalonia/Interop/Win32Interop.cs
nilaoda: @amwx I have used the utility class you provided for the check. However, I found an issue with the function: it returns `false` instead of `true` when the system version exactly matches the major, minor, and build.

amwx (approved):

PR #642, opened 2025-01-24
Fix #641: AppWindow is wrongly resized when changing Fullscreen to Normal mode

### LINKED ISSUE #641: AppWindow is wrongly resized when changing Fullscreen to Normal mode
### Describe the bug
On an `AppWindow`, when setting its `WindowState` to `Fullscreen`, then resetting it to `Normal`, the window doesn't revert the size and position; instead it resizes to the entire screen.

### Lightest Sample Project to demonstrate the issue
[AppWindowFullscreenBug.zip](https://github.com/user-attachments/files/18525021/AppWindowFullscreenBug.zip)

### Desktop/Platform specs
 - OS: Windows 11 24H2
 - AvaloniaUI: 11.2.3 (latest)
 - FluentAvalonia: 2.2.0 (latest)

### Additional context
This issue can be observed in the [attached sample project](https://github.com/user-attachments/files/18525021/AppWindowFullscreenBug.zip) (same as above).

After debugging the code (of the `master` branch), this bug occurs because the `AppWindow` size is saved using the Win32 API after Avalonia has already changed the size by itself. Therefore, the restored size is wrong.

I'm about to create a PR that fixes this bug. Actually, according to my limited tests, only a conditional move of the `OnPropertyChanged` propagation (when the fullscreen is enabled) is needed.
*However, I might be mistaken, as I may not have reviewed all the `AppWindow` breaking changes that could prevent this fix.*

Commode3:
Fix issue #641 by calling the base `OnPropertyChanged` after handling fullscreen changes, when `WindowStateProperty` is concerned on Windows.
I tested this fix using a simple app on Windows 11 24H2, but my tests are likely not exhaustive.

### DIFF
--- a/src/FluentAvalonia/UI/Windowing/AppWindow/AppWindow.cs
+++ b/src/FluentAvalonia/UI/Windowing/AppWindow/AppWindow.cs
@@ -93,7 +93,10 @@ public partial class AppWindow : Window
 
     protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs change)
     {
-        base.OnPropertyChanged(change);
+        if (!this.IsWindows || change.Property != WindowStateProperty)  // Lazy base call OnPropertyChanged
+        {
+            base.OnPropertyChanged(change);
+        }
 
         if (change.Property == IconProperty)
         {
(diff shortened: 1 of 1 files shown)


amwx:
I'm going to go ahead and merge this.  If new issues pop up we'll have to readdress this, but I'm ok with this for now.

Issue #644, opened 2025-01-28
ItemsRepeater leak


Mozart-UA:
**Issue Description:**

It seems that the ItemsRepeater component is causing memory leaks. After some investigation, I have determined that the leaks are due to lingering references to EventHandler in the OnLayoutChanged method.

**Desktop/Platform:**
 - OS: Windows (10 / 11)
 - FluentAvalonia Version (2.1.0 / 2.2.0)
 - Avalonia Version 11.2.3
 
**Steps to Reproduce:**

1. Create a UserControl.
2. Place an empty ItemsRepeater on it, data binding is not necessary.
3. Create a window and display this UserControl in the new window as content.
4. Close the window.

[LeakTestItemsRepeater.zip](https://github.com/user-attachments/files/18578976/LeakTestItemsRepeater.zip)

**Expected Behavior:**
Closing the window should release all references to objects and remove the window from memory.

**Actual Behavior:**
References to the `Window` are held by the `UserControl`, references to the `UserControl` are held by `ItemsRepeater`, and references to `ItemsRepeater` are held by `Layout`. This causes memory leaks and prevents garbage collection. The specific event handlers involved are:

- Layout.MeasureInvalidated -/+ InvalidateMeasureForLayout;
- Layout.ArrangeInvalidated -/+ InvalidateArrangeForLayout;

**Additional context:**
See [attached sample project](https://github.com/user-attachments/files/18578976/LeakTestItemsRepeater.zip). Each time the "Yes Leak" button is pressed, objects will be created, but will never be deleted from memory.
In contrast, clicking the "No Leak" button creates objects, and closing the window marks objects for deletion.

PR #645, opened 2025-01-28
fixes issue: #644

Mozart-UA:
**feat(Core):**
- Add `WeakTypedEventHandler` for binding events with weak references. 

**fix(UI.Controls.ItemsRepeater):**
- Added weak event handlers to `ItemsRepeater` for `InvalidateMeasureForLayout` and `InvalidateArrangeForLayout`. 
- Fixed logical errors in check conditions.
- Improve error message for `DataSource`.

### DIFF
--- a/src/FluentAvalonia/Core/WeakTypedEventHandler.cs
+++ b/src/FluentAvalonia/Core/WeakTypedEventHandler.cs
@@ -1,0 +1,22 @@
+﻿namespace FluentAvalonia.Core;
+public class WeakTypedEventHandler<TSender, TEventArgs> where TEventArgs : EventArgs
+{
+    private readonly WeakReference<TypedEventHandler<TSender, TEventArgs>> _weakHandler;
+
+    public WeakTypedEventHandler(TypedEventHandler<TSender, TEventArgs> handler)
+    {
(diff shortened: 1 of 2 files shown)


amwx:
So I've dug into this a little bit - since no refs to the ItemsRepeater exist outside of that one window in your sample project it should be cleared upon closing the Window, and the event handlers are entirely contained within the ItemsRepeater so once that goes - the events should too.

It looks like this is the culprit:
https://github.com/amwx/FluentAvalonia/blob/3ef356a0919afd9859d0a5d420148d611a45f181/src/FluentAvalonia/UI/Controls/Repeater/ItemsRepeater.props.cs#L30-L31

Everything is being tied back to the `Layout` class, and the defaultValue of the `LayoutProperty` is `new StackLayout()` which is being saved in the property system and that's where the ref is coming from.

I <i>think</i> the correct solution is to take that out, and set the default value in the constructor via `SetCurrentValue()`. That seems to fix the leak as far as I can tell - but it needs additional testing. 

This, of course, wouldn't be an issue if Avalonia would adopt a factory default value like MAUI did but that discussion stalled a long time ago.

PR #654, opened 2025-03-15
Add BreadcrumbBar

### LINKED ISSUE #301: BreadcrumbBar
Are the any plans to port WinUI's [BreadcrumbBar ](https://learn.microsoft.com/en-us/windows/apps/design/controls/breadcrumbbar)to this library?

amwx:
Finally adding one of the older control requests, the `BreadcrumbBar`. This is ported from WinUI with a few tweaks here and there given incompatibilities with Avalonia. Work is still on-going but it is mostly complete for anyone who'd like to test drive the branch.

One thing I am looking to implement is this idea:
https://github.com/microsoft/microsoft-ui-xaml/issues/7213

which is to make the last item in the breadcrumbs interactable or styleable if you choose, as by default in WinUI it is not. 

Closes #301


amwx:
The `BreadcrumbBar` is available in the 2.4-preview 1 package. This control is still in beta

PR #669, opened 2025-05-10
feat: add CloseOnEsc property to allow closing dialogs with the Escape key

### LINKED ISSUE #592: Can I disable Escape key of ContentDialog?
**Is your feature request related to a problem? Please describe.**
No, it's just a small feature.

**Describe the solution you'd like**
Add a property indicating that whether the Escape key is disabled.

**Describe alternatives you've considered**
No.

**Additional context**

xiyaowong:
- Resolve #592


robloo:
This isn't general-purpose enough I think. Also see `IsLightDismissEnabled` on Popup for some related ideas.

That said upstream UWP/WinUI behavior is to invoke the default command and close the dialog when the escape key is pressed. This should simply be the default behavior. We need to find out why there is a request to remove that behavior.

xiyaowong:
> This isn't general-purpose enough I think. Also see IsLightDismissEnabled on Popup for some related ideas.

I'm not very familiar with most parts of this library yet — I just came across this issue and tried to implement it.

> This should simply be the default behavior.

Still the default.

> We need to find out why there is a request to remove that behavior.

This requirement does exist — similar components in other ecosystems also provide such an option. But I understand it's not strictly necessary, since it's possible to cancel the closing by checking whether the result is `None`.

PR #668, opened 2025-04-29
Add check to avoid System.NullReferenceException using NumberBox

ThomasDavenel:
Add check on the string? “Text” of the "_textBox" variable to avoid System.NullReferenceException

Fixes #667

PR #677, opened 2025-06-25
FluentAvalonia v2.4 - Removal of `FAColorPicker`/`ColorPickerButton`

### LINKED ISSUE #670: NumberBox minimum and maximum don't work with coerce
**How to reproduce :**

1. Create a NumberBox with a minimum (for example 0)
2. Write "1" in it
3. Write "-10" in it
4. Press enter
5. NumberBox switch well to 0
6. Rewrite "-10"
7. NumberBox stay in -10

(same behaviour with a maximum)

**Reason :**
The value of the NumberBox is coerce with the function "CoerceValueToRange". 

It works well when you're going out of range for the first time because the value change from the previous one to "0" in this example. But the second time the value stay at 0, so ValueChanged is not triggered and the text is not refresh.

**Desktop/Platform :**
- OS: Windows 11
- FluentAvalonia Version 2.3.0
- Avalonia Version 1.2.8

**Additional context**
Any help to resolve this issue is welcome 😄 
I'll look deeper when I have the time to.

amwx:
v2.4-preview1 has been available for some time, which began with the `BreadcrumbBar` control

One of the bigger things for 2.4 is the removal of the `FAColorPicker` and `ColorPickerButton` controls. As previously noted, these controls have been only partially usable since 11.0, and since I originally wrote these controls Avalonia has gained a color picker of its own, though there are many different features. Right now I don't have the time (or motivation tbh) to rewrite/fix the FAColorPicker so I am removing it altogether - as I'd rather not have incomplete controls around. It may return in the future (especially if requested), but for now it is now gone - please update to using the regular Avalonia ColorPicker.

PR also includes:
- fixes #670, adjusting the coerce function to ensure text gets updated if the Value doesn't change after coersion.
- Changes the default `HorizontalAlignment` on `Expander` to `Stretch` per #663
- #668 
- #673

### DIFF
--- a/src/FluentAvalonia/UI/Controls/NumberBox/NumberBox.properties.cs
+++ b/src/FluentAvalonia/UI/Controls/NumberBox/NumberBox.properties.cs
@@ -123,7 +123,7 @@ public partial class NumberBox
     /// Defines the <see cref="SpinButtonPlacementMode"/> property
     /// </summary>
     public static readonly StyledProperty<NumberBoxSpinButtonPlacementMode> SpinButtonPlacementModeProperty =
-        AvaloniaProperty.Register<NumberBox, NumberBoxSpinButtonPlacementMode>(nameof(SpinButtonPlacementMode), 
+        AvaloniaProperty.Register<NumberBox, NumberBoxSpinButtonPlacementMode>(nameof(SpinButtonPlacementMode),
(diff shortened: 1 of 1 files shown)


robloo:
If there are significant ColorPicker features you think should be included upstream in Avalonia now is the time to open a ticket. We can probably make some control changes for 12.0.

PR #685, opened 2025-08-02
Fix per window DPI awareness in AppWindow

NotroDev:
Hello!

I discovered a serious issue with AppWindow that occurs when a user has two monitors with different scaling settings.

For example, consider this configuration:
- Main display: 1440p at 100% scaling
- Second display: 1080p at 125% scaling

When I launch an app on the second display and it shows a window using AppWindow, the window itself is sized correctly, but the AppWindow content is not - it renders as if using 100% scaling instead of 125%.

Broken AppWindow

The issue can be temporarily resolved by dragging the window to the main monitor and back to the second display, which forces it to render at the correct 125% scale.

If the DPI configuration is reversed, the situation is also reversed: the window gets cut off. This is particularly problematic because having higher scaling on the main display is more common, and when the window is cut off, it can hide critical UI elements including the close button.

It took me several hours to fix this issue as I have little Win32 knowledge, but the final solution seems to be elegant and concise. It works seamlessly without causing visual artifacts like blinking, and I haven't encountered any issues during testing.

### DIFF
--- a/src/FluentAvalonia/Interop/Win32Interop.cs
+++ b/src/FluentAvalonia/Interop/Win32Interop.cs
@@ -40,6 +40,12 @@ internal static unsafe partial class Win32Interop
     [DllImport(s_user32, SetLastError = true)]
     public static extern LRESULT CallWindowProcW(nint lpPrevWndProc, HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam);
 
+    [DllImport("user32.dll", SetLastError = true)]
+    public static extern uint GetDpiForWindow(HWND hwnd);
+    
+    [DllImport("user32.dll", SetLastError = true)]
+    public static extern IntPtr SendMessage(HWND hwnd, uint msg, WPARAM wParam, LPARAM lParam);
+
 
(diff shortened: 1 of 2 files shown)


amwx:
Thank you for wanting to contribute to FA, your contribution and effort is appreciated! In the future for issues like this, please file an issue first with a minimal repro/steps to reproduce. Specifically in this case I'm curious as to why this is happening and would like to fix the root cause rather than slap a hack fix on it, which is where minimal repros are helpful. Do note, however, I do not have a monitor that can go higher than 1920x1080.

1. I'm assuming you've tested this, but if not please do: Does this issue occur with a regular Avalonia window? If yes, then this is an avalonia issue and not FA and and issue will need to be filed upstream. If not, then this is FA specific and we can continue. 

[If FA specific]
2. Can you produce a minimal repro of this (including without a 1440p display). It can be as simple as adding a colored panel to the root of the window, which should still show the issue. This makes it easier for me to reproduce and see what's going on, and I can advise you on the best solution here, if needed.

PR #690, opened 2025-08-12
remove redundant closing braces

SimonCropp:


amwx (approved):
Interesting...Surprised the Xaml compiler doesn't find an issue with this. Thanks!

PR #712, opened 2026-03-08
Update projects to .NET 10, including Android SDK to API 34.

uniflare:
When you updated Desktop etc to .NET 10 the other platforms were missed.
I understand you do not maintain them, but they are still in the repository :).


amwx (approved):
Thanks! You are correct, at least the .net versions should've been updated too.

PR #721, opened 2026-04-15
perf: optimize NumberBoxParser.GetNextNumber performance

textGamex:
Reduce string allocations by using `Span` and leverage `GeneratedRegex` to boost performance.

Benchmark

```C#
    [Benchmark(Baseline = true)]
    public IList<MathToken> Raw()
    {
        return NumberBoxParser.GetTokensRaw("123 + 345 - (1+2)");
    }

    [Benchmark]
    public IList<MathToken> Opt()
    {
        return NumberBoxParser.GetTokens("123 + 345 - (1+2)");
    }
```

| Method | Mean       | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------- |-----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| Raw    | 3,837.8 ns | 73.69 ns | 75.67 ns |  1.00 |    0.03 | 0.6790 |   12784 B |        1.00 |
| Opt    |   232.2 ns |  4.65 ns |  4.56 ns |  0.06 |    0.00 | 0.0291 |     552 B |        0.04 |


amwx (approved):
Thanks, something that I had wanted to do and just forgot about. 
Can you retarget your PR to `master` and update it and then double check that unit tests run successfully. I specifically added some tests for the NumberBox parser with this in mind so I just want to make sure they run without issues. Otherwise, looks good as is.

textGamex:
> Thanks, something that I had wanted to do and just forgot about. Can you retarget your PR to `master` and update it and then double check that unit tests run successfully. I specifically added some tests for the NumberBox parser with this in mind so I just want to make sure they run without issues. Otherwise, looks good as is.

OK, I've switched to the `master` branch and ran the unit tests, and they all passed.

PR #729, opened 2026-04-30
feat: Add IsAotCompatible and add annotations

### LINKED ISSUE #710: Enhance trimming experiences for `Frame.Navigate()` with `[DynamicallyAccessedMembers]`
I tested my app with AOT today and found that it might be helpful to annotate the `sourcePageType` parameter with `[DynamicallyAccessedMembers]`:

https://github.com/amwx/FluentAvalonia/blob/c8f96b4fc0db106e54a359ce71f9b01a20dd1a73/src/FluentAvalonia/UI/Controls/Frame/Frame.cs#L177-L197

The minimal configuration could be `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]`, and this is enough for my project to work. Not sure if `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)]` is neccessary.

OleRoss:
## Description

I always have trim warnings from FluentAvalonia in my applications. This PR adds <IsAotCompatible> and tries to annotate the current code.

## Changes
- Added the IsAotCompatible attribute
- Re-enabled Frame tests (to ensure my changes don't kill functionality)
- Annotated FAFrame
- Used other trim-compatible Enum methods in Color2.cs

## Issues

I have attempted to get full feature parity. However, the FAFrame really relies on reflection and not in an easily fixable way:
1. SetNavigationState is now annotated as `RequiresUnreferencedCode` as we parse types from a string
2. The bigger issue: You can trigger navigation by setting the SourcePageTypeProperty. There is simply no way we can provide trim annotations to the `GetValue(SourcePageTypeProperty, ...)` type. However, I am also not aware of a way to annotate, e.g. the `SourcePageTypeProperty` itself with `RequiresUnreferencedCode`. Therefore, I have decided to suppress the trim warning in the lower code. To notify the user, I added an explanation using `<remarks>` at the relevant properties.

The only way I see to make this provenly trim-friendly would be to provide an IFAFrameFactory with a default implementation that uses reflection, and the option to roll your own factory. That would move the `Activator.CreateInstance` out of the Navigate methods and remove the necessity for any annotations.

Should also fix #710


amwx (approved):

amwx:
Thanks! If you could just retarget to `master` I can merge this for v3 stable. 

Also just a note, if AOT/trimming is really your focus, you should avoid the paths that use `Activator.CreateInstance` and use `Frame.NavigateFromObject()` and a custom `IFANavigationPageFactory` as that allows you to pass viewmodels and create views directly without reflection.

OleRoss:
@amwx Thanks! I retargeted to master. Also, the custom IFANavigationPageFactory makes sense, thx

PR #733, opened 2026-05-12
Fix TabView windowing sample build

KaranocaVe:
## Summary
- populate the existing TabView.TabItems collection instead of assigning the read-only property
- update the windowing sample drag/drop payload to use the current IDataTransfer path
- clean up temporary dragged-tab state when a drag operation completes


amwx:
Thank you for wanting to contribute to FluentAvalonia. However, I am not currently accepting PRs for things within the sample app while v3 is in preview. It will be getting a little renovation for the stable release to ensure the demo pages work. The TabView control was reworked for v3 and I know I'm going to be making changes to that page specifically so I don't want your time and work to be for nothing.

PR #735, opened 2026-05-15
Feat/improve nav item visuals

Qianyiaz:
## Description

This PR makes two changes:

1. Migrates the solution to the `.slnx` format for improved structure and reduced merge conflicts.
2. Adds a rotation animation to the `NavigationViewItem` icon when pressed, for items that use the `setting` style class.
3. Leverage C# field keyword.

## Screenshots

https://github.com/user-attachments/assets/76331c80-fb32-4fcc-9bc6-71bad2d342dd

## Resolved Issues

None


amwx:
Several things here:
- First of all, thank you for wanting to contribute to FA. 
- Please open an issue/feature request prior to PRs. This ensures the following doesn't need to happen
- Please do not do "bulk fix" PRs, ensure the PR has a targeted focus for what it is changing/adding/fixing/etc.

On to this PR:
1. If you would like to isolate the change to .slnx format I would be ok with accepting that PR. I wasn't planning on this as I don't really see the need to, but I won't object. Note that I'm not merging any community PRs during the preview phase of v3 as I don't have the time currently to review & test.

2. A bit of background here. The spinning settings icon that WinUI has comes from their `AnimatedVisual` using a Lottie animation. As Avalonia nor FA has an equivalent we are stuck with the static icon. I have experimented with trying to implement some sort of AnimatedVisual-like system but there's a lot of complications that I still need to think through. If I'm going to add animations like this to the core FA library, I'd like it to take an approach like this. Your solution works fine for yourself but I don't think it is a good general purpose solution, so unfortunately I'm going to decline this change.

3. The final change to use the new C# `field` keyword will also not be accepted. I really, really hate how C# is starting to hide a lot of details behind the compiler - it doesn't make for cleaner, more understandable code in my opinion. I put FA on .net10 to ensure we get perf benefits in the runtime and to avoid `#if` branches, not to adopt the new C# syntax that adds little to no value.

If you would still like to submit the solution file change, either change the title of the PR & revert everything but that or close this PR and open a new one.

Qianyiaz:
Thanks for your good jobs.

PR #747, opened 2026-06-24
Decouple FAFrame BackStack from CacheSize and prevent memory leaks

chenjt2001:
## Description

This PR removes the incorrect `BackStack` clearing logic in `FAFrame` when navigating in `FANavigationMode.New`. 

https://github.com/amwx/FluentAvalonia/blob/246e09f17e5538ca4a434a87e2676d343b8be6e0/src/FluentAvalonia/UI/Controls/Frame/FAFrame.cs#L511-L525

Previously, the code removed entries from the `BackStack` when its count reached the `CacheSize`. This logic conflated the concept of navigation history with page caching. This PR fixes the issue based on the following reasons:

1. `CacheSize` is meant to control how many page instances are kept alive in memory, not the navigation depth. Whether a user can navigate back is controlled by `IsNavigationStackEnabled`. Even if a page is evicted from the cache, it should still remain in the `BackStack` so it can be recreated upon backward navigation.
2. Truncating the `BackStack` based on cache limits prevents deep backward navigation. This behavior is inconsistent with the standard Microsoft WinUI `Frame` behavior.
3. Throughout the rest of the codebase, `CacheSize` is correctly paired with `_pageCache.Count` to manage instances. Memory management and eviction should strictly operate on `_pageCache`, leaving the logical `BackStack` intact.
4. The previous code had a logical flaw when `CacheSize` was explicitly set to `0`. On the first navigation, both `BackStack.Count` and `CacheSize` were `0`, bypassing the removal block but executing `BackStack.Add(prevEntry)`. On subsequent navigations, `BackStack.Count` would be greater than `CacheSize` (e.g., 1 != 0), meaning `BackStack.RemoveAt(0)` would never execute. Ironically, setting `CacheSize` to `0` resulted in an infinite `BackStack`.

By removing the `if (BackStack.Count == CacheSize)` block, the navigation history functions as expected, while `_pageCache` continues to handle the actual page instance memory limits.

[The `Instance` property of the `entry` added to the BackStack is set to `null`](https://github.com/chenjt2001/FluentAvalonia/blob/0d333563685abf885f723610388c4a761dfd8f0c/src/FluentAvalonia/UI/Controls/Frame/FAFrame.cs#L505-L514) to release resources; therefore, saving the entry does not cause a memory leak.

In addition, this PR addresses several other issues related to FAFrame; please refer to the commit messages and code comments for details.

### DIFF
--- a/src/FluentAvalonia/UI/Controls/Frame/FAFrame.cs
+++ b/src/FluentAvalonia/UI/Controls/Frame/FAFrame.cs
@@ -465,7 +465,15 @@ public partial class FAFrame : ContentControl
             if (!wasPageSet)
             {
                 // Check if we already have an instance of the page in the cache
-                entry.Instance = CheckCacheAndGetPage(entry.SourcePageType);
+                // Context will not be null if NavigateCore is called from GoBack/GoForward and the entry was created from NavigateFromObject
+                if (entry.Context != null)
+                {
+                    entry.Instance = CheckCacheAndGetPage(null, entry.Context);
+                }
+                else
+                {
+                    entry.Instance = CheckCacheAndGetPage(entry.SourcePageType, null);
+                }
             }
 
             if (entry.Instance == null)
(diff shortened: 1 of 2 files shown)


amwx (approved):
Thanks. Actually ran in to something that was related to this recently, so perfect timing :)

Issue #757, opened 2026-07-18
ViewManager: Debug assertion fails in PrunePinnedElements due to incorrect Owner state when pinned elements are cleared


lfeng1420:
**Describe the bug**
When a pinned element (e.g., an element with keyboard focus) is cleared, a `Debug.Assert` fails in `ViewManager.PrunePinnedElements()` at L288:
```csharp
Debug.Assert(vi.Owner == VirtualizationInfo.ElementOwner.PinnedPool);
```

**Root Cause**
1. When an element gains focus, `ViewManager.UpdateFocusedElement()` calls `UpdatePin(element, true)`, which increments `_pinCounter` via `VirtualizationInfo.AddPin()`.
2. When this pinned element is later cleared, `ViewManager.ClearElementToPinnedPool()` is invoked. Since `virtInfo.IsPinned` is true, the element is added to `_pinnedPool`.
3. The bug: After adding the element to `_pinnedPool`, the method calls:
    ```csharp
    virtInfo.MoveOwnershipToLayoutFromPinnedPool();
    ```
    This sets the element's Owner to Layout instead of PinnedPool.
4. When `PrunePinnedElements()` iterates through `_pinnedPool`, it asserts that each element's `Owner` should be `PinnedPool`, causing the assertion to fail.

**Proposed Fix**
In `ViewManager.ClearElementToPinnedPool()` (at L834), change `virtInfo.MoveOwnershipToLayoutFromPinnedPool()` to `virtInfo.MoveOwnershipToPinnedPool()`.

**Desktop/Platform (please complete the following information):**
 - OS: Windows 11
 - FluentAvalonia Version 3.0.2
 - Avalonia Version 12.0.4

amwx:
Good catch - a typo on my part and your fix matches WinUI correctly. I'll accept a PR if you're willing.

lfeng1420:
> Good catch - a typo on my part and your fix matches WinUI correctly. I'll accept a PR if you're willing.

Thanks! PR is up: #759

PR #759, opened 2026-07-26
fix(ViewManager): debug assertion failure in PrunePinnedElements

lfeng1420:
## Description
This PR fixes a debug assertion failure in `PrunePinnedElements`.

In `ClearElementToPinnedPool`, the method incorrectly called `MoveOwnershipToLayoutFromPinnedPool()` instead of  
`MoveOwnershipToPinnedPool()`. This was a simple typo that caused elements returned to the pinned pool to have the incorrect `ElementOwner` state, triggering an assertion when `PrunePinnedElements` ran.

The fix aligns the behavior with WinUI and ensures the element ownership transition logic is correct. No behavioral or visual changes are introduced.

## Screenshots
N/A – no visual changes.

## Resolved Issues
#757


amwx (approved):
Thanks!

PR #760, opened 2026-07-27
fix(focus): scope initial focus searches to their intended root

maggch97:
`FAContentDialog.SetupDialog` and `FATabView` both ask for an initial focus target like this:

```csharp
var mgr = TopLevel.GetTopLevel(this).FocusManager;
// TODO: v3 - does this work?
var next = mgr.FindNextElement(NavigationDirection.Next, new FindNextElementOptions { SearchRoot = this });
next?.Focus();
```

Answering that `TODO`: no, it does not. Per the WinUI contract Avalonia mirrors, `FindNextElementOptions` is only honoured for directional navigation and is silently ignored for tab navigation. From the `FocusManager.FindNextElement` docs:

> Tab navigation (`FocusNavigationDirection.Previous` and `FocusNavigationDirection.Next`)
> cannot be used with `FindNextElementOptions`. Only directional navigation (`Up`, `Down`,
> `Left`, or `Right`) is valid.

Avalonia implements exactly that: for `Next`/`Previous` it uses `options?.FocusedElement ?? Current` and never reads `SearchRoot`. So neither call was scoped to anything - both started from whatever held focus in the window and walked the entire visual tree.

## Consequences

1. The chosen element could be *outside* the dialog / the new tab - the opposite of what the surrounding comments say the code is for.

2. Worse, it can hang the application outright. Avalonia's tab stop walk fails to terminate when the starting element is nested inside a container with `KeyboardNavigationMode.Once`: `GetNextTabStop` resets its parent walk back to the focused element's parent on every `Once` hit and oscillates between two nodes forever. Since `NavigationView` items live in exactly such a container, opening a `ContentDialog` while focus sat on a nested navigation item froze the UI thread at 100% CPU permanently. That is how this was found, in a production app.

## The fix

Both sites switch to `FocusManager.FindFirstFocusableElement(root)`, which is correctly scoped and is the operation actually wanted - the first focusable element *inside* the given root, rather than the tab stop following some unrelated element. It also takes a different code path in Avalonia, so the hang goes away without waiting for the upstream fix (verified below).

Two incidental improvements in `FAContentDialog`:

- The `else Focus()` branch now actually implements the fallback its comment promises ("focus the ContentDialog itself to pull focus away from the main visual tree"). The old `next?.Focus()` did nothing at all when `next` was null.
- `TopLevel.GetTopLevel(this)` is no longer dereferenced without a null check.

## The other two call sites are fine

For completeness, the remaining uses of these APIs were checked and left alone:

- `FANavigationView.cs:2004` passes `NavigationDirection.Up`, i.e. directional navigation, where `SearchRoot` *is* honoured. Correct as written.
- `FABreadcrumbBar.GetFindNextElementOptions()` has no callers.

## Verification

- New regression test `DialogWithNoDefaultButtonDoesNotHangWhenFocusIsInsideAOnceContainer` in `ContentDialogTests`, showing a dialog while focus is parked inside a nested `KeyboardNavigationMode.Once` container.
- Confirmed it catches the bug: with the `FAContentDialog` change reverted, and against the unmodified Avalonia 12.0.0 package, the test never completes - the run had to be killed after 95s of CPU time. With the change it passes in under a second.
- Full `FluentAvaloniaTests` suite: 168 tests, 0 failed (150 passed, 18 pre-existing skips). The `FATabView` change is covered only by the existing 30 TabView tests; I did not find a clean way to drive its focus-moving branch from a headless test.

## Things I deliberately did not change

- **The timing of the search.** When `SetupDialog` runs (from `DialogLoaded`) the dialog's content is not yet effectively visible, so `FindFirstFocusableElement` returns null and the fallback is what actually takes effect. That is still a strict improvement over walking into the window behind the dialog, but if the intent is to focus the first control of the user's content, the search likely needs to move to after `Opened`. That is a behavioural decision for you rather than something to fold into a hang fix. The same reasoning applies to the stale "it is necessary to call UpdateLayout here" comment in `FATabView`.
- **`Focusable` on `FAContentDialog`.** It is not set, so the `Focus()` fallback does not currently succeed either and focus simply stays put. Setting it would make the dialog itself a tab stop, which has its own implications - flagging rather than deciding.
- **The `Timeout` on the new test.** It is there for consistency with neighbouring tests, but an xUnit timeout cannot pre-empt a CPU-bound loop on the UI thread: if this regresses, CI will hang rather than report a failed assertion.

## Description

A clear and concise description of the PR. Why was the PR necessary and how does it solve the issue?

## Screenshots

Screenshots or a screen recording of the visual changes associated with this PR. This is not required, but may be helpful to me to see the changes without needing to build your branch. Of course if your change doesn't change any visuals, then this is not needed.

## Resolved Issues

If this PR closes any active issues, please link them here

### DIFF
--- a/src/FluentAvalonia/UI/Controls/ContentDialog/FAContentDialog.cs
+++ b/src/FluentAvalonia/UI/Controls/ContentDialog/FAContentDialog.cs
@@ -434,10 +434,19 @@ public partial class FAContentDialog : ContentControl, ICustomKeyboardNavigation
                 // ContentDialog itself to pull focus away from the main visual tree so weird things don't happen
                 // The latter shouldn't happen in 99% of cases as either something in the user content will be able
                 // to take focus OR there should always be at least one button which can take focus
-                var mgr = TopLevel.GetTopLevel(this).FocusManager;
-                // TODO: v3 - does this work?
-                var next = mgr.FindNextElement(NavigationDirection.Next, new FindNextElementOptions { SearchRoot = this });
-                next?.Focus();
+                //
+                // Do NOT go back to FindNextElement(NavigationDirection.Next, new FindNextElementOptions
+                // { SearchRoot = this }) here. Per the WinUI contract Avalonia mirrors, FindNextElementOptions
+                // is only honoured for directional navigation and is silently ignored for Next/Previous, so
+                // SearchRoot scoped nothing: the search started from whatever held focus outside the dialog and
+                // walked the whole window. FindFirstFocusableElement is both correctly scoped and the operation
+                // actually wanted here - the first focusable element inside the dialog, not the tab stop
+                // following some unrelated element.
+                var next = FocusManager.FindFirstFocusableElement(this);
+                if (next is not null)
+                    next.Focus();
+                else
+                    Focus();
 
 #if DEBUG
                 Logger.TryGet(LogEventLevel.Debug, "ContentDialog")?.Log("SetupDialog", "Set initial focus to {next}", next);
--- a/src/FluentAvalonia/UI/Controls/TabView/FATabView.cs
+++ b/src/FluentAvalonia/UI/Controls/TabView/FATabView.cs
@@ -878,9 +878,14 @@ public partial class FATabView : TemplatedControl
                 
                 if (shouldMoveFocusToNewTab)
                 {
-                    var focusable = TopLevel.GetTopLevel(this)?.FocusManager?.FindNextElement(NavigationDirection.Next,
-                        new FindNextElementOptions { SearchRoot = _tabContentPresenter });
-                    
+                    // FindNextElement(Next, ...) cannot be scoped: FindNextElementOptions is only
+                    // honoured for directional navigation, so SearchRoot was ignored here and the
+                    // search ran from whatever held focus anywhere in the window. Besides picking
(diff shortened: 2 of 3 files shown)


amwx:
I will accept this change, however
1. The change should also be applied to `TaskDialog` as that is essentially the same as `ContentDialog`
https://github.com/amwx/FluentAvalonia/blob/af36ebd3011c9a5cf8eaa735d65727b2071beaf7/src/FluentAvalonia/UI/Controls/TaskDialog/FATaskDialog.cs#L609-L619

And more importantly,
I am going to ask that you resubmit the PR without Claude or any other LLM as a coauthor. I have updated the contributing guidelines to include the AI statement that was previously only in the PR description box (which was my mistake) - which requires all code to be human written (or at least don't make it apparent AI wrote the code). I don't want any AI agent in the contributor/commit history on this repo. Thanks for understanding.

Also to address the points in the "did not change section"
1. This is by design, uses who want to customize focus on load behavior should handle the dialog `Opened` event and set focus there. To ensure nothing gets in the way of user handling, FA attempts focus at the earliest possible point.
2. Also by design
3. I'm assuming claude wrote this, but I'm not sure what its point is here. The timeout is there b/c the dialogs require async code and if something goes wrong I don't want a test to hang forever, and that's how xUnit force ends the test in this scenario.

Thanks for your contribution. Closing PR per request above.

